PR #84, opened 2024-08-06
feat(crd): streamline instrumentation configuration via custom resource

basti1302:
Instead of three boolean flags, we now use an enumeration with instrumentation modes.


basti1302:
waiting for https://linear.app/dash0/issue/ENG-2181/streamline-crd-configs-and-rename-to-dash0monitoring#comment-06148467 to get resolved

PR #92, opened 2024-08-13
feat: automatically collect logs of monitored applications

mmanciop:
This PR is the first iteration of supporting ootb log collection for monitored applications (i.e., logs from pods having the `dash0.com/instrumented=true` label).


basti1302:
https://github.com/dash0hq/dash0/pull/3143 is about storing and processing `dash0.monitoring.*` resource attributes correctly, instead of discarding them.

test-resources/otlp-sink/otlp-sink.yaml
```diff
@@ -60,7 +65,7 @@ metadata:
   labels:
     app.kubernetes.io/name: deployment
     app.kubernetes.io/component: manager
-    dash0.com/enable: "false"
+    dash0.com/instrumented: "false"
```
basti1302, line 68: @mmanciop last review/looking over the changes before merging. I don't understand this particular change. 

It probably does not matter, but `dash0.com/instrumented` is typically something that the operator would set. I would not expect that it is set externally. I think the previous choice `dash0.com/enable: "false"` made a little more sense. In the end, neither label should have an effect and we could probably delete it altogether, since there is never a Dash0Monitoring resource in the otlp-sink namespace anyway.

images/filelogoffsetsynch/src/filelogoffsetsynch.go
```diff
@@ -0,0 +1,336 @@
...
+	FilelogOffsetDirectoryPath string
+}
+
+type patch struct {
+	BinaryData map[string]string `json:"binaryData,omitempty"`
+}
+
+// TODO Add support for sending_queue on separate exporter
+// TODO Set up compaction
+// TODO Set up metrics & logs
```
basti1302, line 44: @mmanciop a couple of TODOs left in the code here, are they meant to stay? Delete? Convert to Linear issues?

internal/backendconnection/otelcolresources/config.yaml.template
```diff
@@ -1,115 +1,238 @@
...
-                endpoint: ${env:MY_POD_IP}:4317
-            http:
-                endpoint: ${env:MY_POD_IP}:4318
+  otlp:
+    protocols:
+      grpc:
+        endpoint: ${env:MY_POD_IP}:4317
+      http:
+        endpoint: ${env:MY_POD_IP}:4318
+  # TODO Turn on conditionally for monitored namespaces
```
basti1302, line 87: Same here, is this TODO something that needs to be taken care of? The way it is now I think the TODO would end up in config maps in the customer's systems, which is not good for optics IMO.

basti1302 (approved):

PR #155, opened 2024-09-25
feat(injector): add custom Dash0 LD_PRELOAD injector

mmanciop:
Add to the Dash0 instrumentation image an `LD_PRELOAD` injector that:
1) Works irrespective of the libc flavour used by the dynamically-linked host process (e.g., Node v8, CPython, the JVM)
2) Modified on-demand the value of environment variables looked up by the host process through the `getenv` LibC API (currently implemented: `NODE_OPTIONS`)

The injector replicates internally a subset of the libc APIs related with string comparison and manipulation. That code is adapted from muse libc, and the respective copyright notice is included.


mmanciop:
TODOs as of f292f128cb2a2921197d9f4bead577f7b2454752:

1. Integrate multi-arch image building and testing
2. Test more scenarios relevant for injection, e.g., unrelated var remain untouched
3. Adjust the env vars set by the operator on injection to use `LD_PRELOAD`

PR #156, opened 2024-09-25
feat: manage Perses dashboards via the operator

basti1302:


basti1302:
cc @mdonkers for awareness :)

PR #192, opened 2024-11-01
Untested change to add dataset for operator self reporting

RichardoC:
Unable to test this due to docker desktop refusing to work from locations with IPv6


basti1302:
Hey Richard,

thanks for contributing this, it's a great addition. Having the dataset configurable via Helm definitely makes sense. I'll pick this PR up today or tomorrow 👍

RichardoC:
Replaced by #198

basti1302:
I would have reported back here but you were faster than me :)  I needed to extend the changes a bit, which is why I took your commit and created a follow-up PR from it. Again, thanks for the contribution.

PR #215, opened 2024-12-02
feat(injector): External resource detection for OpenTelemetry-instrumented applications

mmanciop:


mmanciop:
Well I honestly did not expect the PR to go green on the first attempt.

images/instrumentation/injector/src/dash0_injector.c
```diff
@@ -85,28 +88,70 @@ char val2[1012];
...
     char *origValue = __getenv(name);
+	char *podUid = __getenv(DASH0_POD_UID_ENV_VAR_NAME);
+	char *containerName = __getenv(DASH0_POD_CONTAINER_NAME_VAR_NAME);
     int l = __strlen(name);
 
+	char *otelResourceAttributesVarName = OTEL_RESOURCE_ATTRIBUTES_ENV_VAR_NAME;
     char *nodeOptionsVarName = NODE_OPTIONS_ENV_VAR_NAME;
-    if (__strcmp(name, nodeOptionsVarName) == 0)
-    {
+	if (__strcmp(name, otelResourceAttributesVarName) == 0)
```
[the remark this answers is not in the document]
mmanciop: I tired `clang-format -i` and I am reasonably pleased with the outcome. (I don't actually care how exactly it looks, as long as it is consistent.)

images/instrumentation/injector/src/dash0_injector.c
```diff
@@ -82,31 +94,85 @@ char *__getenv(const char *name)
...
+          __strcat(val1, ",");
+        }
+
+        __strcat(val1, origValue);
+      }
     }
 
-    return origValue;
-}
+    return val1;
```
basti1302, line 158: _question / double check / rambling:_ According to https://opentelemetry.io/docs/specs/otel/resource/sdk/#specifying-resource-information-via-an-environment-variable all characters outside of "baggage-octet range MUST be percent-encoded."

https://www.w3.org/TR/baggage/#definition:

> baggage-octet          =  %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
>                          ; US-ASCII characters excluding CTLs,
>                          ; whitespace, DQUOTE, comma, semicolon,
>                          ; and backslash

Some SDKs will actually validate the string and crash the process if there are invalid characters:
https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-resources/src/detectors/EnvDetectorSync.ts#L118-L120 ->
https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-resources/src/detectors/EnvDetectorSync.ts#L133-L135

We basically add user provided input here to the env var. We either need to be _very_ sure that none of the sources (pod name, container name, ...) can contain invalid characters or do percent-encoding in the injector.

* namespace names, pod names & container names: I think all of those should fall under the following and thus they should be fine: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names -> lowercase alphanumeric characters, '-' or '.' See also: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L177
* pod uids: obviously those are fine

I'm bringing it up just in case, would you agree with this being safe?

images/instrumentation/injector/test/scripts/run-tests-within-container.sh
```diff
@@ -42,11 +42,12 @@ run_test_case() {
...
   expected=$3
   existing_node_options_value=${4:-}
+  existing_otel_resource_attributes=${5:-}
   set +e
   if [ "$existing_node_options_value" != "" ]; then
-    test_output=$(LD_PRELOAD="$injector_binary" TEST_VAR=value NODE_OPTIONS="$existing_node_options_value" node index.js "$command")
+    test_output=$(LD_PRELOAD="$injector_binary" TEST_VAR=value DASH0_NAMESPACE_NAME=my-namespace DASH0_POD_NAME=my-pod DASH0_POD_UID=275ecb36-5aa8-4c2a-9c47-d8bb681b9aff DASH0_CONTAINER_NAME=test-app NODE_OPTIONS="$existing_node_options_value" OTEL_RESOURCE_ATTRIBUTES="${existing_otel_resource_attributes}" node index.js "$command")
   else
-    test_output=$(LD_PRELOAD="$injector_binary" TEST_VAR=value node index.js "$command")
+    test_output=$(LD_PRELOAD="$injector_binary" TEST_VAR=value DASH0_NAMESPACE_NAME=my-namespace DASH0_POD_NAME=my-pod DASH0_POD_UID=275ecb36-5aa8-4c2a-9c47-d8bb681b9aff DASH0_CONTAINER_NAME=test-app OTEL_RESOURCE_ATTRIBUTES="${existing_otel_resource_attributes}" node index.js "$command")
```
basti1302, line 50: _nitpick:_ Even with passing an empty string for argument 5, the test case does not accurately reflect the reality of running without OTEL_RESOURCE_ATTRIBUTES being set, because setting an env var to an empty string and not having it set at all is not the same. If you don't mind I will extend this with another check like we already have with `if [ "$existing_node_options_value" != "" ]; then`.

images/instrumentation/injector/src/dash0_injector.c
mmanciop: This is an excellent point, we should make it safe.

images/instrumentation/injector/test/scripts/run-tests-within-container.sh
basti1302: done

images/instrumentation/injector/src/dash0_injector.c
basti1302: After double checking, we concluded that the character set of K8s object names is indeed more restrictive than the baggage-octet, so the current code is safe.

basti1302 (approved):

PR #220, opened 2024-12-16
feat: make kubeletstatsreceiver work ootb on Minikube and Docker Desktop

mmanciop:
The Kubelet stats endpoint on Minikube, Docker Desktop and Kind appears to use self-signed certificates that are not signed by the cluster's certificate authority. This means that they are invalid for the Daemonset's collector, preventing kubelet-based metrics from being collected, and degrading the first user experience of the operator for end users that try it in a local cluster.

This change works around it by disabling TLS veritification for the kubeletstats receiver if the collector is running on a node named "minikube" or "docker-desktop", and documenting ways to solve the issue for Kind or `minikube -p <profile_name>`.


basti1302 (approved):
LGTM, thanks!

PR #225, opened 2024-12-21
feat: maintain views as code

bripkens:
# Why
Sharing views with your team mates is very helpful to standardize workflows. Maintaining these views as code just helps go the extra mile.

# What
Support the `Dash0View` custom resource and maintain is it within Dash0 via the API.


config/crd/bases/operator.dash0.com_dash0monitorings.yaml
```diff
@@ -248,6 +248,13 @@ spec:
                   See https://github.com/dash0hq/dash0-operator/blob/main/helm-chart/dash0-operator/README.md#managing-dash0-check-rules-with-the-operator
                   for details. This setting is optional, it defaults to true.
                 type: boolean
+              synchronizeViews:
+                default: true
+                description: |-
+                  If enabled, the operator will watch Dash0 view resources in this namespace and create corresponding views in Dash0 via the Dash0 API.
+                  See https://github.com/dash0hq/dash0-operator/blob/main/helm-chart/dash0-operator/README.md#managing-dash0-check-rules-with-the-operator
```
basti1302, line 255: issue: wrong link (points to check rules management)

```suggestion
                  See https://github.com/dash0hq/dash0-operator/blob/main/helm-chart/dash0-operator/README.md#managing-dash0-views
```

helm-chart/dash0-operator/README.md
```diff
@@ -875,6 +882,48 @@ Prometheus Rule Synchronization Results:
...
+* The operator will only pick up Dash0 view resources in namespaces that have a Dash0 monitoring resource
+  deployed.
+* The operator will not synchronize Dash0 view resources in namespaces where the Dash0 monitoring resource
+  has the setting `synchronizeViews` set to `false`. (This setting is optional and defaults to `true` when
+  omitted.)
+
+Furthermore, the custom resource definition for Dash0 views needs to be installed in the cluster. There are two
+ways to achieve this:
+
+**TODO**
```
basti1302, line 903: suggestion: Since this wouldn't be a "foreign" resource type (like Prometheus rules or Perses dashboards) but a **Dash0** resource type, we should rather create a Dash0View resource type in this repo (similar to Dash0Monitoring or Dash0OperatorConfiguration) and install it unconditionally via the operator Helm chart. You already added the Helm bits for that. We also need the corresponding Go type definitions in `api/dash0monitoring/v1alpha/dash0view_types.go`, similar to the existing `_types.go` files we already have there for our existing two Dash0 resource types.

Then we can remove this section, this particular precondition would always be satisfied when the operator is installed.

Unless, of course, there is an open-source, de-facto standard K8s resource type describing views, that roughly matches what we need. But I don't think that something like this exists.

helm-chart/dash0-operator/templates/operator/custom-resource-definition-dash0view.yaml
```diff
@@ -0,0 +1,181 @@
+---
```
basti1302, line 1: remark: I assume you used one of the other Helm custom resource types as a blueprint. The way we usually would create and maintain the Helm definitions for custom resources would be Go-code first:
- Describe the resource type in `api/dash0monitoring/v1alpha/dash0view_types.go`
- Run `make generate` to have `config/crd/bases/operator.dash0.com_dash0views.yaml` generated
- Copy that file over to `helm-chart/dash0-operator/templates/operator/custom-resource-definition-dash0view.yaml`

It is a bit cumbersome (due to the fact that the operator framework and tooling works with kustomize instead of Helm), but it ensures that actual instances of the resource type can be properly deserialized at runtime.

helm-chart/dash0-operator/templates/operator/custom-resource-definition-monitoring.yaml
```diff
@@ -397,6 +405,34 @@ spec:
                 description: Shows results of synchronizing Prometheus rule resources
                   in this namespace via the Dash0 API.
                 type: object
+              viewSynchronizationResults:
```
basti1302, line 408: suggestion: This change in the monitoring resource should also follow the Go-code first approach outlined above.

basti1302 (commented):
I know this is still a draft but I was curious and thought I might as well leave some early feedback while having a look. Of course I can also take over this effort, depending on what you prefer.

PR #252, opened 2025-01-31
feat: restore k8s.replicaset.* metrics by keeping only non-zero values

mmanciop:


internal/backendconnection/otelcolresources/deployment.config.yaml.template
```diff
@@ -91,9 +98,9 @@ receivers:
       k8s.namespace.phase:
         enabled: false
       k8s.replicaset.desired:
-        enabled: false
+        enabled: true
       k8s.replicaset.available:
-        enabled: false
+        enabled: true
```
basti1302, line 103: nitpick: there is no need to set them to true explicitly, we should just remove the four lines 100-103 entirely, as these two metrics are enabled by default.

basti1302 (approved):

internal/backendconnection/otelcolresources/deployment.config.yaml.template
basti1302: fixed

PR #269, opened 2025-02-07
Various features

mmanciop:
1. Remove unnecessary resources update when dealing with finalizers
2. Add support for setting `service.{name|namespace|version|instance.id}` via the `app.kubernetes.io/*` labels
3. Add support for setting additional resource attributes via the `resource.opentelemetry.io/<key>: <value>` annotations
4. Collect `container.image.{name|tag}` through the `k8sattributeprocessor`


helm-chart/dash0-operator/README.md
```diff
@@ -727,6 +727,36 @@ resource attribute auto-detection.
...
+The labels and annotations specified below will be ignored when set, for example, in the metadata of deployments and daemonsets, as opposed  to the metadata of their respective pod templates.
+
+The following [standard Kubernetes labels](https://kubernetes.io/docs/reference/labels-annotations-taints/#labels-annotations-and-taints-used-on-api-objects) are mapped toi resource attributes as follows:
+
+* `app.kubernetes.io/name` => `service.name`
+* `app.kubernetes.io/version` => `service.version`
+* `app.kubernetes.io/part-of` => `service.namespace`
+* `app.kubernetes.io/instance` => `service.instance.id`
+
+Note: the `OTEL_SERVICE_NAME` environment variable will have precedence over the `app.kubernetes.io/name` label.
```
basti1302, line 742: I imagine pre-set OTEL_RESOURCE_ATTRIBUTES take precendence as well?

helm-chart/dash0-operator/README.md
```diff
@@ -727,6 +727,36 @@ resource attribute auto-detection.
...
+
+Additionally, the any annotation in in the form of `resource.opentelemetry.io/<key>: <value>` will be mapped to the `<key>=<value>` resource attribute.
```
basti1302, line 744:
```suggestion
Additionally, any annotation in in the form of `resource.opentelemetry.io/<key>: <value>` will be mapped to the `<key>=<value>` resource attribute.
```

helm-chart/dash0-operator/README.md
```diff
@@ -727,6 +727,36 @@ resource attribute auto-detection.
 This allows Dash0 users to avoid the hassle of manually adding the OpenTelemetry SDK to their applications.
 Dash0 takes care of it automatically.
 
+### Specifying additional resource attributes via pod labels and annotations
```
basti1302, line 730: I don't think this section should be here in the "This section provides a quick behind-the-scenes glimpse into how the Dash0 operator's workload instrumentation for tracing works" part of the README. 

The fact that we implement this configurability via the injector is a technical detail, the option to set the resource attributes are an important feature though. It needs to be in the "Configuration" main section, starting in line 126. There is no good sub-section to add it there yet, we should probably create a new sub-section called something like "Per-workload configuration" or something similar and put this new content there, alongside with what we have now in "Disabling Auto-Instrumentation for Specific Workloads".

internal/workloads/workload_modifier.go
```diff
@@ -200,22 +205,22 @@ func (m *ResourceModifier) ModifyStatefulSet(statefulSet *appsv1.StatefulSet) Mo
 	return m.modifyResource(&statefulSet.Spec.Template, &statefulSet.ObjectMeta)
 }
 
-func (m *ResourceModifier) modifyResource(podTemplateSpec *corev1.PodTemplateSpec, meta *metav1.ObjectMeta) ModificationResult {
-	if hasBeenModified := m.modifyPodSpec(&podTemplateSpec.Spec); !hasBeenModified {
+func (m *ResourceModifier) modifyResource(podTemplateSpec *corev1.PodTemplateSpec, podMeta *metav1.ObjectMeta) ModificationResult {
```
basti1302, line 208: issue: this rename is not correct, the second parameter is the meta from the workload type, e.g. if this is called from `ModifyDeployment` it will be the deployment's meta, not the pod meta. Please revert.

basti1302 (changes requested):
Looks good, but I think there is some disparity between what we say in the updated docs and how the code actually works (deployment labels/annotations vs pod labels/annotations).

helm-chart/dash0-operator/README.md
mmanciop: yes

basti1302: To clarify: The intention of my comment was that we should also spell that out explicitly here.

mmanciop: Done in a4cc1d3

helm-chart/dash0-operator/README.md
mmanciop: Done in a4cc1d3

basti1302 (approved):

PR #308, opened 2025-03-11
fix: add fallback __environ symbol

basti1302:


images/instrumentation/injector/src/dash0_injector.c
```diff
@@ -89,6 +93,12 @@ int __strncmp(const char *_l, const char *_r, size_t n) {
 }
 
 char *__getenv(const char *name) {
+  // If we cannot link the __environ symbol,
+  // always return that env var is not set.
+  if (__environ == NULL) {
+    return 0;
+  }
```
basti1302, line 100: q: Just to double check, and for my understanding: Could that possibly lead to a situation where a binary that does not link as we expect is not able to read environment variables at all, because we always return 0 here for any __getenv call? Or do we assume that if __environ is not defined, that also our __getenv override will never be actually called?

mmanciop: Well, my thinking is the following: there is no LibC flavor I ever found that exposes a global symbol `getenv`, but not `__environ`: it's a such hardcoded convention that it's how `/proc/self/environ` works afaict. And if this convention breaks, the only chance you should have to get into trouble is if you weak-link a `getenv` symbol that does something else entirely. Whoever manages to do that, I want to hear about it, because I would have so many questions.

basti1302: Sounds good, that's what I figured. I will try to get an additional test in place that is compiled similar to `aws-vpn-cni`, reproduces the `undefined symbol` crash we see currently with the injector from `main`, and then check how it behaves with this injector fix in place.

basti1302: The test I added in the second commit shows that this is not an issue. 👍

basti1302:
This is stale and has been superseded by https://github.com/dash0hq/dash0-operator/pull/356 and https://github.com/dash0hq/dash0-operator/pull/359.

PR #401, opened 2025-06-02
ci: use depot runners

basti1302:


basti1302:
Closing this unmerged since the depot runners seem to be a bit slower than the default runners.

The main build&test task usually takes around less than 3:00 to maybe 3:30 minutes with `ubuntu-latest`  (example [here](https://github.com/dash0hq/dash0-operator/actions/runs/15349833036/job/43195001380) and [here](https://github.com/dash0hq/dash0-operator/actions/runs/15374201109/job/43257133675)) but it took around 5 minutes with `depot-ubuntu-22.04`  (examples [here](https://github.com/dash0hq/dash0-operator/actions/runs/15386311742/job/43285572350), [here](https://github.com/dash0hq/dash0-operator/actions/runs/15387203260/job/43288303350) and [here](https://github.com/dash0hq/dash0-operator/actions/runs/15386909732/job/43287377705), the latter one didn’t even finish because golangci-lint was so slow it ran into a timeout).

Thus staying with `ubuntu-latest` for now.

PR #466, opened 2025-08-04
feat(deployment): custom kube rbac image option

Jacse:
For users (like us) that need to deploy the helm chart behind a firewall we need to be able to define custom images.

Added tests and otherwise tried to keep to conventions I've seen.


basti1302:
Hey @Jacse, thanks for the contribution. Sorry for not getting back to you earlier, I just came back from vacation today. The changes all look pretty good, thanks so much for including tests etc. I'll see that I can merge and release this later this week.

PR #490, opened 2025-09-02
Remove link-time dependency from __environ and setenv

mmanciop:
The injector may be linked to binaries that, while dynamically linked, do not link a LibC flavor. Since the injector requires the linker the `__environ` and `setenv` symbols at linking time, and those cannot be satisfied due to the absence of a LibC, the injector effectively prevents the program from starting due to a linker error like the following:

> Defaulted container "aws-node" out of: aws-node, aws-eks-nodeagent, aws-vpc-cni-init (init), dash0-instrumentation (init)
> /app/aws-vpc-cni: symbol lookup error: /__dash0__/dash0_injector.so: undefined symbol: __environ
    
This change introduces a way to look up at runtime, in an initialization step, the location in the process' memory of the `__environ` and `setenv` symbols (if any exist!), so that we can use them to read and manipulate the process environment without depending on them at link time.
    
The way we do this is in three steps:
1. Find out which LibC the process has (if any) -- We already did this read Elf metadata (`DT_NEEDED`) as some OpenTelemetry SDKs need to be injected differently (using different binaries) depending on the LibC flavor.
2. In the loaded LibC, find the location of the `dlsym` function by reading Elf metadata. To accomplish this, the change "forks" the `ElfDynLib` from Zig's `dynamic_library` and modifies it to remove stuff we don’t care about (e.g., symbol version matching), and to prevent it from memory mapping (as what we need was already mapped in memory by the process loading); think of it as “just enough Elf to find a symbol the process already has”
3. Let the LibC's `dlsym` handle the fifty quintillion corner cases (aliases, lazy loading, linker resolution order, `__environ`) exported by the main binary, etc.) and find us `__environ` and `setenv`


mmanciop:
@basti1302 I forgot to turn on by default the static tests

basti1302:
I think this is finally ready to be merged. 🎉 💥

PR #501, opened 2025-09-15
ci: remove potentially outdated version tags

basti1302:
In bda19c006bc91bdc85ff1ca294ffafa48055ece6, we replaced all semver tags of third-party Github actions with their SHAs. This was done to avoid having Github implicitly using a more recent version of an action, and instead requiring an explicit code change to update a third-party GH action. This was motivated by security incidents like https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/

In the commit, the GH action SHAs were added together with a version comment at the end. However, the GH action that updates the SHAs automatically does not update the comment, so with PRs like https://github.com/dash0hq/dash0-operator/pull/496/files the comment becomes outdated and hence misleading.


ten15bit (approved):
👍

PR #508, opened 2025-09-17
feat(startup, chart): improved ready check & support for helm ... --wait

basti1302:
Previously, the ready check of the operator manager pod was basically a no-op, marking the operator as ready immediately after startup. This could lead to issues when trying to install custom resource directly after running helm install, since the validation webhook service was not reachable yet. The new improved ready check waits until the webhook service endpoint has a port assigned.

Another issue that can prevent deploying a Dash0Monitoring resource directly after running helm install is the following: When asking the Dash0 operator Helm chart to create a Dash0OperatorConfiguration resource via operator.dash0Export, this happens in an eventually- consistent fashion. That is, the resource is created at startup by the operator manager pod. Even with the improved readiness check mentioned above, the operator configuration resource might not have been created and/or reconciled when trying to deploy the Dash0Monitoring resource directly after helm install finishes. This would make the validation webhook for the monitoring resource fail, because having an operator configuration resource with export settings is a prerequisite. To resolve this, there is now a new post-install Helm hook that waits until the operator configuration resource has been created. This cannot be included in the operator manager pod's readiness check, since it will only happen after the operator manager pod has marked itself as ready.

In summary: When installing the operator into a cluster with helm install --wait, users can now be sure that they can deploy a Dash0Monitoring resource successfully immediately after helm install --wait has finished.


internal/controller/operator_configuration_controller.go
```diff
@@ -330,8 +330,9 @@ func (r *OperatorConfigurationReconciler) applyApiAccessSettings(
 
 		}
 	} else {
-		logger.Info("The API endpoint setting required for managing dashboards or check rules via the operator is " +
-			"missing or has been removed, the operator will not update dashboards nor check rules in Dash0.")
+		logger.Info("The API endpoint setting required for managing dashboards, check rules, synthetic checks and	 " +
+			"views via the operator is missing or has been removed, the operator will not update these resources " +
+			"in Dash0.")
```
basti1302, line 335: ^ this is unrelated to the readiness check change. Managing synthetic checks and views with the operator has recently been implemented and these messages were not up to date yet.

internal/controller/synthetic_check_controller_test.go
```diff
@@ -37,23 +37,9 @@ const (
...
-}
-
-func (l *leaderElectionAwareMock) NeedLeaderElection() bool {
-	return true
-}
-
-func (l *leaderElectionAwareMock) IsLeader() bool {
-	return l.isLeader
-}
-
```
basti1302, line 56: Moved this into test/util.

internal/predelete/operator_pre_delete_handler.go
```diff
@@ -59,55 +59,55 @@ func NewOperatorPreDeleteHandlerFromConfig(config *rest.Config) (*OperatorPreDel
 	}, nil
 }
 
-func (r *OperatorPreDeleteHandler) SetTimeout(timeout time.Duration) {
-	r.timeout = timeout
+func (h *OperatorPreDeleteHandler) setTimeout(timeout time.Duration) {
```
basti1302, line 62: No actual changes in the pre-delete handler, just some minor refactorings.

internal/startup/auto_operator_configuration_handler.go
```diff
@@ -70,22 +80,38 @@ func (r *AutoOperatorConfigurationResourceHandler) CreateOrUpdateOperatorConfigu
...
+				"resource",
+		)
+		// Block until NotifiyOperatorManagerJustBecameLeader has been called.
+		select {
+		case <-ctx.Done():
+			logger.Error(ctx.Err(), "context cancelled while waiting for this replica to become leader")
+			return
+		case <-r.hasBecomeLeaderChan:
+			logger.Info("this replica has become leader, proceeding with creating/updating the operator configuration resource")
+		}
```
basti1302, line 96: ^ this is also only tangentially related to the new readiness check -- previously, all replicas would try to create/update the operator configuration resource, now this depends on leader election, so only one pod tries to do that.

internal/startup/auto_operator_configuration_handler.go
```diff
@@ -118,60 +144,6 @@ func (r *AutoOperatorConfigurationResourceHandler) validateOperatorConfiguration
 	return nil
 }
 
-func (r *AutoOperatorConfigurationResourceHandler) waitForWebserviceEndpoint(
-	ctx context.Context,
-	logger *logr.Logger,
-) error {
```
basti1302, line 124: The new readiness check shares most of the logic for waiting for the web service endpoint, so this has been moved to ready_check.go.

test-resources/bin/util
```diff
@@ -447,51 +447,15 @@ run_helm() {
   helm_install_command+=" ${OPERATOR_HELM_CHART:-helm-chart/dash0-operator}"
   echo Helm install command:
   echo "$helm_install_command"
-  $helm_install_command
-
-  wait_for_operator_manager_and_webhook
```
basti1302, line 452: All of the logic in wait_for_operator_manager_and_webhook is no longer necessary, because that is now handled by the extended readiness check and the Helm post-install hook. (Because we are now also passing `--wait` to `helm install`.)

test-resources/bin/util
```diff
@@ -500,22 +464,22 @@ deploy_otlp_sink_if_requested() {
...
+  touch "$otlp_sink_dir/traces.jsonl"
+  touch "$otlp_sink_dir/metrics.jsonl"
+  touch "$otlp_sink_dir/logs.jsonl"
+
+  kubectl apply -f "$tmpfile"
+  kubectl rollout status \
+    deployment \
+    otlp-sink \
+    --namespace otlp-sink \
+    --timeout 1m
```
basti1302, line 482: Unrelated, just cleaning up some formatting issues here.

test/e2e/dash0_operator_configuration_resource.go
```diff
@@ -87,28 +86,6 @@ func deployDash0OperatorConfigurationResource(
 	}
 }
 
-func waitForAutoOperatorConfigurationResourceToBecomeAvailable() {
```
basti1302, line 90: There are no new e2e tests, but by removing these wait conditions (and `waitForManagerPodAndWebhookToStart`), a lot of the existing e2e test cases implicitly verify that a Dash0 monitoring resource can be deployed directly after running `helm install --wait`.

ten15bit (approved):
looks good 👍

PR #510, opened 2025-09-18
docs(chart): recommend a filelog offset volume for OPA and AKS

basti1302:
The Open Policy Agent gatekeeper (and in turn AKS with the Azure Policy add-on enabled) have the same issue as the Kyverno admission controller, that is, they frequent ConfigMap updates of the operator cause high CPU usage for these components; making the use of a persistent volume for filelog offsets more or less mandatory.


ten15bit (approved):
👍

PR #512, opened 2025-09-18
feat(exporter): add support for insecureSkipVerify

ten15bit:
This PR adds support for `insecureSkipVerify` in the exporter config to enable the usage of self-signed certificates.

### DIFF
--- a/api/operator/common/common_shared_types.go
+++ b/api/operator/common/common_shared_types.go
@@ -135,7 +135,7 @@ type GrpcConfiguration struct {
 	// +kubebuilder:validation:Optional
 	Headers []Header `json:"headers,omitempty"`
 
-	// Explicitly defines whether TLS is used. Per default, a secure connection is assumed, unless the Endpoint starts with 'http://'.
+	// Explicitly defines whether TLS is used. Per default, a secure connection is assumed, unless the endpoint starts with 'http://'.
 	//
 	// +kubebuilder:validation:Optional
 	Insecure *bool `json:"insecure,omitempty"`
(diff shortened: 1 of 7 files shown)


internal/webhooks/monitoring_validation_webhook.go
```diff
@@ -150,6 +150,16 @@ func (h *MonitoringValidationWebhookHandler) validateExport(
...
 					"operator configuration does not have an export configuration either."), true
 		}
+	} else {
+		if monitoringResource.Spec.Export.Grpc != nil {
+			if util.ReadBoolPointerWithDefault(monitoringResource.Spec.Export.Grpc.Insecure, false) &&
+				util.ReadBoolPointerWithDefault(monitoringResource.Spec.Export.Grpc.InsecureSkipVerify, false) {
+				return admission.Denied(
+					"The provided Dash0 monitoring resource has both insecure and insecureSkipVerify " +
+						"explicitly enabled for the GRPC export. This is an invalid combination. " +
+						"Please either set export.grpc.insecure=true or export.grpc.insecureSkipVerify=true."), true
```
basti1302, line 160: **nitpick:** To me this sounds like it could be read as "You have to set at least one the flags to true." Maybe something along the lines of

```suggestion
						"Please set at most one of these two flags to true."), true
```

ten15bit: updated the error message in a separate commit since it also affected the operator config validation webhook

ten15bit:
@basti1302 thanks for the review! :)

PR #521, opened 2025-09-24
ci: add dependabot config for docker

ten15bit:
Small PR to add automatic version updates for Docker.

Note: I didn't add a config for Helm, since it only checks the chart dependencies and not the images used in the templates/manifests. Since we currently don't depend on any other charts, that check wouldn't provide any value.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -46,11 +46,11 @@ updates:
       - "/images/collector"
       - "/images/configreloader"
       - "/images/filelogoffsetsync"
-      # - "/images/filelogoffsetvolumeownership"
+      - "/images/filelogoffsetvolumeownership"
       - "/images/instrumentation"
     schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -39,6 +39,20 @@ updates:
...
       interval: "daily"
 
+  # Docker
+  - package-ecosystem: "docker"
+    directories:
+      - "/"
+      - "/images/collector"
+      - "/images/configreloader"
+      - "/images/filelogoffsetsync"
+      # - "/images/filelogoffsetvolumeownership"
```
ten15bit, line 49: waiting for https://github.com/dash0hq/dash0-operator/pull/520 to be merged

basti1302: It's merged now. ✅

basti1302 (approved):

PR #545, opened 2025-10-07
chore: remove rbac markers and outdated generated files

ten15bit:
The RBAC-related config has been managed directly in the Helm chart [for a while](https://github.com/dash0hq/dash0-operator/commit/c093e0e7212f3e620cadde014b93a1dfec5c4c4f) and the present RBAC markers and generated files are unused and outdated.

This PR removes the markers and generated files and adds a note to clarify where the RBAC-related config can be found.

The decision to manually manage those files and not make use of code generation was based on the following reasons:
- it is clearer to manage the permission in a single place than having several markers scattered across the controllers package
- the generated manifests have to be manually copied into the Helm chart anyway

### DIFF
--- a/config/rbac/README.md
+++ b/config/rbac/README.md
@@ -1,3 +1,3 @@
 # Note about RBAC
 
-We manually manage RBAC-related configuration directly in the Helm chart.
+We manually manage RBAC-related configuration directly in the [dash0-operator Helm chart](../../helm-chart/dash0-operator).


config/rbac/README.md
```diff
@@ -0,0 +1,3 @@
+# Note about RBAC
+
+We manually manage RBAC-related configuration directly in the Helm chart.
```
basti1302, line 3: _thought_: should we provide a pointer to `helm-chart/dash0-operator/templates/operator/cluster-roles.yaml` here?

basti1302 (approved):
Thanks for cleaning this up!

config/rbac/README.md
ten15bit: link added :link:

basti1302: Well the actual permissions are really mostly listed in cluster-roles.yaml only, I think. But just linking to the chart is good, I agree 👍

PR #560, opened 2025-10-13
feat(chart): rudimentary support for GKE autopilot

basti1302:
Add a Helm parameter for GKE autopilot specific deployment settings.

When enabled, the following changes are applied:
- omit default dash0.com/enable node affinities
- add additional namespaces to the NotIn namespaceSelectors to all three mutating webhooks; GKE autopilot would add these namespaces anyway
- disable the extra metadata labels container.id and k8s.volume.type for kubeletstats receiver metrics, collecting these requires access to the /pod endpoint of the kubelet API which is not available in GKE autopilot
- disable all four utilization metrics for kubeletstats receiver metrics, collecting these requires access to the /pod endpoint of the kubelet API which is not available in GKE autopilot:
  - k8s.pod.cpu_limit_utilization
  - k8s.pod.cpu_request_utilization
  - k8s.pod.memory_limit_utilization
  - k8s.pod.memory_request_utilization
- disable the host metrics receiver, it requires mounting the full host file system as a volume, which is not permitted on GKE autopilot

It is currently unclear whether we also need to disable host ports. According to the documentation (see for example
https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-security: "Setting specific host ports is disallowed, ..."), and in earlier experiments this showed up as an issue. However, in the most recent tests, using host ports for the OpenTelemetry collector daemonset and using them in the OTEL_EXPORTER_OTLP_ENDPOINT setting on workloads worked, despite what the GKE autopilot documentation says.

### DIFF
--- a/helm-chart/dash0-operator/templates/operator/post-install-hook.yaml
+++ b/helm-chart/dash0-operator/templates/operator/post-install-hook.yaml
@@ -9,7 +9,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     app.kubernetes.io/name: dash0-operator
-    app.kubernetes.io/component: uninstallation-process
+    app.kubernetes.io/component: installation-process
     {{- include "dash0-operator.labels" . | nindent 4 }}
     dash0.com/enable: "false"
   annotations:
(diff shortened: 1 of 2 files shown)


helm-chart/dash0-operator/README.md
```diff
@@ -2050,15 +2093,6 @@ spec:
...
 
-## Notes on Azure AKS
-
-In [AKS](https://azure.microsoft.com/products/kubernetes-service) clusters that have the
-[Azure Policy add-on](https://learn.microsoft.com/azure/aks/use-azure-policy) enabled, it is highly recommended to
-[use a volume for filelog offsets](#providing-a-filelog-offset-volume)
-instead of the default filelog offset config map.
-Using the default config map filelog offset storage in AKS clusters with this add-on can lead to severe performance
-issues.
-
```
basti1302: This section did not actually change, I only moved it up so all cloud-vendor specific things are in one place.

PR #572, opened 2025-10-27
test(e2e): various improvements 

basti1302:
* chore: clean up Makefile and CONTRIBUTING.md
* chore: do not auto-install pre-push hook with random make target
* test(e2e): remove all "docker build" calls from end-to-end tests, use make targets instead
* chore: do not try to guess whether an image should be built
* chore(Makefile): add IMAGE_PLATFORMS, rename image targets
* test(e2e): support IMAGE_REPOSITORY_PREFIX, IMAGE_TAG and PULL_POLICY
* chore(test-resources): add SKIP_IMAGE_BUILDS

----

This PR bundles a couple of changes around image builds, e2e tests and test scripts. The intention is to clean up some awkward heuristics from the Makefile and the e2e tests and to get a couple steps closer to be able to easily run e2e tests against remote clusters, and also with a broader variety of local setups (Docker Desktop vs. Podman etc.).

There are still more things to do, but I guess this PR has gotten too big already 🫤 

Out of scope for this PR, but TBD for follow-up PRs:
- should we remove the facilities to run `kind load docker-image ...` to make the images available on kind cluster nodes, and instead only support pulling images from a registry (i.e. a local kind registry, a remote registry, ...)
- we should probably remove the heuristics to auto-detect whether the current cluster is a kind cluster
- add make targets for pushing to a registry after building, both for operator images and test app images
- we might want to get rid of any assumptions about image name prefixes, in particular in container_images.go#localImages etc.
- we might want to switch the default for imagePullPolicy=Never in the e2e tests, which only works with Docker Desktop, and instead leave the imagePullPolicy unset by default?


test/e2e/applications_under_test.go
```diff
@@ -146,41 +146,12 @@ var (
...
 	Expect(
 		runAndIgnoreOutput(
 			exec.Command(
-				"docker",
-				"build",
-				applicationPath,
-				"-t",
-				imageName,
+				"make",
+				"test-app-images",
```
basti1302: Image builds for the actual operator images (operator-controller, instrumentation, collector, ...) are no longer triggered by the e2e tests. Instead, they need to have been built before and need to be present or pullable from the target cluster.

However, building auxiliary images like test applications and the Dash0 API mock still are triggered from within the e2e tests --  for now. We can discuss whether we want to remove this make call in a follow up PR, or whether we want to make it optional via an environment variable switch. It will come up again once we move towards running the e2e tests in a remote cluster, where the test app images also need to be present in some registry that the cluster can pull from.

Makefile
```diff
@@ -247,134 +234,130 @@ instrumentation-test-lint: npm-installed
...
 ##@ Build
 
 .PHONY: build
-build: husky-setup-hooks manifests generate fmt vet ## Build manager binary.
+build: manifests generate fmt vet ## Build manager binary.
 	go build -o bin/manager cmd/main.go
 
-.PHONY: run
-run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./cmd/main.go
```
basti1302: I removed this target for starting the operator directly on your local machine (without deploying it as an image to a cluster) as well, I think it is rather pointless to do that.

Makefile
```diff
@@ -247,134 +234,130 @@ instrumentation-test-lint: npm-installed
...
 $(eval $@_CONTEXT = $(3))
 $(eval $@_DOCKERFILE = $(4))
-if [[ -n "$($@_IMAGE_REPOSITORY)" ]]; then                                                                \
-  if [[ "$($@_IMAGE_REPOSITORY)" = *"/"* && "$($@_IMAGE_REPOSITORY)" != *"/library/"* ]]; then                                                         \
-    echo "not rebuilding the image $($@_IMAGE_REPOSITORY), this looks like a remote image";               \
-  else                                                                                                    \
-    dockerfile=$($@_DOCKERFILE);                                                                          \
-    if [[ -z $$dockerfile ]]; then                                                                        \
-        dockerfile=$($@_CONTEXT)/Dockerfile;                                                              \
-    fi;                                                                                                   \
```
basti1302: This is the part of the PR to stop applying weird heuristics about whether to build an image or not.

CONTRIBUTING.md
```diff
@@ -67,117 +70,30 @@ make test
...
+everything they deploy in their `AfterAll`/`AfterEach` hooks.
 
-The e2e tests might sometimes not be the best tool to troubleshoot the operator, simply because they remove everything
-they deploy in their `AfterAll`/`AfterEach` hooks.
-The scripts in `test-resources/bin` can be used for creating a specific scenario and inspecting then its behavior.
+The scripts in `test-resources/bin` can be used for creating a specific scenario with one shell command, and then
+inspecting its behavior.
 They are also useful to actually report data to an actual Dash0 backend.
 
 If you haven't created `test-resources/.env` yet, copy the file `test-resources/.env.template` to `test-resources/.env`.
```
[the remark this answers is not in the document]
basti1302: Goog point, I rephrased that now. 👍

PR #579, opened 2025-10-31
feat: add custom labels to synthetic check definitions

yiquanzhou:
Allow adding custom labels to synthetic check resource definitions in `spec.labels` field. These labels can be used for routing notifications when a synthetic check fails. 

image


basti1302 (approved):
LGTM! 🎉

PR #582, opened 2025-11-03
chore: add GitHub issue templates

basti1302:


basti1302:
> nitpick: file names are a bit inconsistent atm (`-` vs `_`) 🙂

Good catch. I blame GitHub, since I used the settings UI dialog to set up these templates 😬 

image

Anyway, fixed now.

PR #588, opened 2025-11-04
fix: recheck string length after dropping registry part

ten15bit:
The current code would panic with `runtime error: slice bounds out of range` when the string length is < 63 chars after dropping the registry part. We should only truncate the remaining string if it's necessary.


basti1302 (approved):

ten15bit:
> Good catch.
> 
> nitpick: Should this be a `fix: ` commit or a `chore: ` when it lands in main?

changed it to `fix`

PR #590, opened 2025-11-06
feat(injector): introduce new log levels, default to error, respect immediately

basti1302:
feat(injector): introduce new log levels "warn" and "error" in addition to "debug", "info" and "none".

* All error messages are now printed at level error.
* The default log level is now error instead of info, that is, the
  success message "environment injection finished" is no longer printed
  by default. This is especially important for scenarios where the
  stderr stream of the binary is piped into another process for parsing,
  the previous behavior would break the parsing.

---

fix(injector): respect DASH0_INJECTOR_LOG_LEVEL immediately at startup

Previously, DASH0_INJECTOR_LOG_LEVEL would be ignored until the
injector's startup routine had called updateStdOsEnviron(), which only
happened after figuring the libc details. This is far too late to
suppress log messages entirely with DASH0_INJECTOR_LOG_LEVEL=none.

Now the injector will read DASH0_INJECTOR_LOG_LEVEL immediately at
startup by reading /proc/self/environ, instead of waiting for the
pointer to libc's `__environ` to be discovered.

Additionally, this commit removes support for interpreting
DASH0_INJECTOR_DEBUG, in favor of DASH0_INJECTOR_LOG_LEVEL=debug


images/instrumentation/injector/src/auxv.zig
```diff
@@ -8,7 +8,7 @@ const print = @import("print.zig");
 // Executables requiring getauxval will bind to libc's getauxval, not the symbol exported here.
 pub export fn getauxval(auxv_type: u32) callconv(.C) usize {
     var auxv_file = std.fs.openFileAbsolute("/proc/self/auxv", .{}) catch |err| {
-        print.printMessage("Failed to open /proc/self/auxv: {}", .{err});
```
basti1302: The injector previously did not differentiate between log levels besides debug, info and none, so every error message was printed basically at level info. Now, with more log levels available, every error message changed from printMessage to printError, which makes the diff a bit noisy.

basti1302:
@claude review

PR #593, opened 2025-11-10
test: convert test app manifests to Helm charts

basti1302:
Each test application (Node.js, JVM, .NET, Python) now has its own Helm
chart, and instead of deploying them via fixed Kubernetes manifests,
they are now deployed via these Helm charts, both in the e2e tests as
well as in the semi-manual test scenarios.

This allows to more easily template the test application workloads, in
particular the image name and pull policy.


basti1302:
@claude review

PR #600, opened 2025-11-13
test(e2e): check matching telemetry via in-cluster service

basti1302:


test/e2e/otlp-sink/helm-chart/templates/collector-config-map.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: v1
```
basti1302: This is basically unchanged, I only split test-resources/otlp-sink/otlp-sink.yaml into separate Helm templates (config map, service deployment etc.)

Dockerfile
```diff
@@ -7,9 +7,11 @@ WORKDIR /workspace
 COPY go.mod go.mod
 COPY go.sum go.sum
 
-# This particular COPY needs to be executed before go mod download since it is referenced by a replace directive
-# in go.mod.
+# These particular COPY instructions need to be executed before go mod download since it is referenced by a replace
+# directive in go.mod.
 COPY images/pkg/common/ images/pkg/common/
+# Only used in test/e2e sources, not in production code.
+COPY test/e2e/pkg/shared test/e2e/pkg/shared
```
basti1302: Yay/nay? This (together with `replace github.com/dash0hq/dash0-operator/test/e2e/pkg/shared => ./test/e2e/pkg/shared` in the main go.mod`) allows to reuse some code between the telemetry-matcher and the e2e test suite (some shared types and string constants, like query param names) Those would otherwise need to be duplicated.

The downside is that a Go package that is meant for testing (test/e2e/pkg/shared) is now available in the Go build for the operator-controller image. Obviously, it is not referenced in production code, but this still feels a bit icky.

IMO this is an acceptable tradeoff, but it's probably debatable. WDYT?

ten15bit: ➕ for having the e2e tests in a separate module in the long term

the current solution is fine for now though (imho) 👍

ten15bit (approved):

test/e2e/otlp_sink.go
```diff
@@ -5,113 +5,113 @@ package e2e
...
+func init() {
+	// disable keep-alive
+	t := http.DefaultTransport.(*http.Transport).Clone()
+	t.DisableKeepAlives = true
+	telemetryMatcherHttpClient = &http.Client{Transport: t}
+}
 
-		return nil
-	}()).To(Succeed())
+func determineTelemetryMatcherImage() {
```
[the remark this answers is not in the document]
basti1302: > The next iteration (which I talked about earlier today, but have still not pushed :-/ )

pushed now -> https://github.com/dash0hq/dash0-operator/pull/605

PR #602, opened 2025-11-14
test: use kind with port-mapping for test scripts

ten15bit:
## The main changes in this PR are

- adds a port mapping to the worker node in the kind-config
- adds a label to this node so the nginx controller can be scheduled on the right node
- patches the nginx deployment with a `preferredDuringSchedulingIgnoredDuringExecution` using the label from above
- changes the (default) service type of the test apps from `LoadBalancer` to `ClusterIP`, since the apps will be accessible via the nginx ingress

## Minor changes/fixes:

- the Helm templates for the test apps used `port` instead of `targetPort` - since those are currently always the same in our test setup, this hasn't caused any issues so far, but it would fail when using different ports
- `helm install`ing the jvm and .net apps didn't work because of a missing `\`
- the new python test app was missing in the `test-app-images` target in the makefile

## How to test

After creating the cluster and running the test scenario as usual (e.g. setting the required env vars and running `create_cluster_and_registry.sh` and `test-scenario-01-aum-operator-cr.sh`), the test application can be reached via nginx on the port defined by the port-mapping (`8080`) and the path defined in the ingress resource. E.g.:

- `curl localhost:8080/deployment/nodejs/dash0-k8s-operator-test`
- `curl localhost:8080/deployment/python/dash0-k8s-operator-test`
-...

### DIFF
--- a/test-resources/nginx/kustomization.yaml
+++ b/test-resources/nginx/kustomization.yaml
@@ -1,0 +1,27 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - https://kind.sigs.k8s.io/examples/ingress/deploy-ingress-nginx.yaml
+
+patches:
+  - target:
+      kind: Deployment
+      name: ingress-nginx-controller
+    patch: |-
+      apiVersion: apps/v1
+      kind: Deployment
+      metadata:
+        name: ingress-nginx-controller
+      spec:
+        template:
+          spec:
+            affinity:
+              nodeAffinity:
+                preferredDuringSchedulingIgnoredDuringExecution:
+                  - weight: 100
+                    preference:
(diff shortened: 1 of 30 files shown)


test-resources/nginx/kustomization.yaml
```diff
@@ -0,0 +1,27 @@
...
+      apiVersion: apps/v1
+      kind: Deployment
+      metadata:
+        name: ingress-nginx-controller
+      spec:
+        template:
+          spec:
+            affinity:
+              nodeAffinity:
+                preferredDuringSchedulingIgnoredDuringExecution:
```
ten15bit: this should ensure that nginx will still be scheduled when running in a cluster that doesn't use a port-mapping and therefore doesn't have a node with the `nginx-ingress=true` label (e.g. docker-desktop, EKS)

PR #604, opened 2025-11-16
feat: add support for Prometheus CRDs

ten15bit:
Adds support for Prometheus CRDs (`ServiceMonitor`, `PodMonitor`, `ScrapeConfig`).

## Main changes

When a `Dash0Configuration` is configured with

```yaml
prometheusCrdSupport:
    enabled: true
```

and there is at least one namespace with a `Dash0Monitoring` resource having `prometheusScraping.enabled` set to `true`, the operator will deploy the [OpenTelemetry Target Allocator](https://github.com/open-telemetry/opentelemetry-operator/tree/main/cmd/otel-allocator) and configure it to watch Prometheus CRs in the namespaces with scraping enabled.

Additionally, the operator will update the [prometheusreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/README.md) in the Dash0 daemonset collectors to query the allocator for scrape targets.

## Additional changes

- refactorings to reduce duplication between the collector and target-allocator controllers
- some improvements/additions to the test scripts

### DIFF
--- a/test-resources/bin/lib/util
+++ b/test-resources/bin/lib/util
@@ -833,7 +833,6 @@ install_third_party_crds() {
   install_prometheus_crds
 }
 
-# todo: integrate with version check
 install_prometheus_crds() {
   helm upgrade --install \
     --namespace "$target_namespace" \
--- a/test/e2e/targetallocator_resources.go
+++ b/test/e2e/targetallocator_resources.go
@@ -105,5 +105,5 @@ func verifyThatTargetAllocatorIsNotDeployed(operatorNamespace string) {
 				targetAllocatorName,
 			))).ToNot(Succeed())
 	}
-	Eventually(verifyTargetAllocatorDeploymentDoesNotExist, 60*time.Second, time.Second).Should(Succeed())
+	Consistently(verifyTargetAllocatorDeploymentDoesNotExist, 60*time.Second, time.Second).Should(Succeed())
 }
--- a/internal/controller/monitoring_controller.go
+++ b/internal/controller/monitoring_controller.go
@@ -157,6 +157,9 @@ func (r *MonitoringReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		if err = r.reconcileOpenTelemetryCollector(ctx, nil, &logger); err != nil {
 			return ctrl.Result{}, err
 		}
+		if err = r.reconcileOpenTelemetryTargetAllocator(ctx, &logger); err != nil {
+			return ctrl.Result{}, err
+		}
 
 		return ctrl.Result{}, nil
 	} else if checkResourceResult.StopReconcile {
@@ -207,6 +210,9 @@ func (r *MonitoringReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		// being true. We still want to handle this case correctly for good measure (reconcile the otel collector, then
 		// stop the reconcile of the monitoring resource).
 		if r.reconcileOpenTelemetryCollector(ctx, monitoringResource, &logger) != nil {
+			return ctrl.Result{}, err
+		}
+		if r.reconcileOpenTelemetryTargetAllocator(ctx, &logger) != nil {
 			return ctrl.Result{}, err
 		}
 		// The Dash0 monitoring resource is slated for deletion, the finalizer has already been removed in the last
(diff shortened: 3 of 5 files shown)


test-resources/bin/lib/util
```diff
@@ -807,9 +827,32 @@ helm_install_test_application() {
   $helm_install_command
 }
 
+helm_install_podinfo() {
+  helm install \
+    --namespace "$target_namespace" \
+    --values test-resources/podinfo/values.yaml \
+    podinfo \
+    oci://ghcr.io/stefanprodan/charts/podinfo
+}
```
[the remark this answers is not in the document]
ten15bit: yes, will go over it before the PR is ready

test/e2e/operator.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const (
 	localHelmChart            = "helm-chart/dash0-operator"
-	operatorHelmReleaseName   = "e2e-tests-operator-helm-release"
+	operatorHelmReleaseName   = "e2e-tests-operator-hr"
```
ten15bit: combining the release name with the target-allocator deployment name resulted in a string >=64 chars

helm-chart/dash0-operator/values.yaml
```diff
@@ -290,6 +298,16 @@ operator:
     # override the default image pull policy
     pullPolicy:
 
+  # the container image to use for the target-allocator component (there should usually be no reason to override this)
+  targetAllocatorImage:
+    repository: "ghcr.io/open-telemetry/opentelemetry-operator/target-allocator"
+    # overrides the image tag, which defaults to the chart appVersion.
+    tag: "0.140.0"
```
basti1302: _Non-blocking:_ As a follow-up we probably should make sure we create PRs to update the version regularly. Either via Dependabot or by integrating it into  `.github/workflows/update-collector-components.yaml` or something similar.

basti1302: +1 for separate PR, definitely.

test/e2e/targetallocator_resources.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+var (
+	targetAllocatorName = fmt.Sprintf(
+		"%s-opentelemetry-target-allocator-deployment",
+		operatorHelmReleaseName,
+	)
+)
+
+func deployPrometheusCrds(cleanupSteps *neccessaryCleanupSteps) {
```
basti1302: _question:_ In the test-resources script, the Prometheus CRDs are installed via `helm upgrade ... prometheus-community/prometheus-operator-crds`. What is the reason for doing it differently in the e2e tests?

test-resources/bin/lib/util
```diff
@@ -810,6 +830,21 @@ helm_install_test_application() {
 install_third_party_crds() {
   kubectl apply --server-side -f https://raw.githubusercontent.com/perses/perses-operator/refs/tags/v0.2.0/config/crd/bases/perses.dev_persesdashboards.yaml
   kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.86.2/example/prometheus-operator-crd/monitoring.coreos.com_prometheusrules.yaml
+  install_prometheus_crds
+}
+
+# todo: integrate with version check
```
basti1302: q: which version check is this referring to?

test-resources/bin/test-cleanup.sh
```diff
@@ -59,6 +59,8 @@ if [[ "$wait_for_third_party_resource_deletion" = "true" ]]; then
   sleep 2
 fi
 
+helm uninstall --namespace "$target_namespace" prometheus-crds --ignore-not-found || true
```
basti1302: 👍

Makefile
```diff
@@ -171,8 +173,12 @@ build-all-push-all-test-e2e: all-images push-all-images test-e2e ## Builds and p
 # flag), and ginkgo's own timeout.
 .PHONY: test-e2e
 test-e2e: ## Run the end-to-end tests. When testing local code, container images should be built beforehand (or use target build-images-test-e2e).
+ifdef GINKGO_FOCUS
+	go run github.com/onsi/ginkgo/v2/ginkgo -v -focus="$(GINKGO_FOCUS)" test/e2e
```
basti1302: You are probably aware of it, but just in case: you can also simply change a specific `Describe`, `DescribeTable` or `It` to `FDescribe`, `FDescribeTable` or `FIt` temporarily to focus on a particular suite or test case. I find this often more convenient compared to providing -focus. Similar with `XDescribe`, `XIt` to exclude test cases.

Either way, having support for -focus in the Makefile is definitely a good addition. 👍

test/e2e/targetallocator_resources.go
basti1302: Ha, good point. I didn't remember that this is also different between scripts and e2e tests. I also don't see a real reason why it should be this way. But kudos for keeping it consistent with what is there already. I'm fine with leaving it like this for the new CRDs. Eventually we might want to align it between scripts and e2e tests both for the "new" Prometheus CRDs and also for PrometheusRule and PersesDashboard, but it's also not exactly a high priority. Let's leave it like this for now. 👍

PR #605, opened 2025-11-17
test(e2e): align e2e tests with latest updates from test-resources/bin

basti1302:


test-resources/bin/lib/constants
```diff
@@ -23,7 +23,7 @@ local_reg_port='5001'
 
 # defaults
 default_kind_cluster="dash0-operator-lab"
-default_kind_config="test-resources/kind-config-lr.yaml"
```
basti1302: I removed the old test-resources/kind-config.yaml and renamed the new test-resources/kind-config-lr.yaml to test-resources/kind-config.yaml.

test-resources/bin/lib/util
```diff
@@ -747,25 +747,29 @@ build_test_application() {
...
-  echo "building container image for application under monitoring ($image_name)"
-  docker build . -t "$image_name"
-  popd > /dev/null
+  local make_rule="$1"
+  TEST_APP_DOTNET_IMAGE_REPOSITORY="$test_app_dotnet_image_repository" \
+    TEST_APP_DOTNET_IMAGE_TAG="$test_app_dotnet_image_tag" \
+    TEST_APP_JVM_IMAGE_REPOSITORY="$test_app_jvm_image_repository" \
+    TEST_APP_JVM_IMAGE_TAG="$test_app_jvm_image_tag" \
+    TEST_APP_NODEJS_IMAGE_REPOSITORY="$test_app_nodejs_image_repository" \
+    TEST_APP_NODEJS_IMAGE_TAG="$test_app_nodejs_image_tag" \
```
basti1302: The need to list all these env vars whenever calling make is quite ugly, maybe we can get rid of it later. For now I think it is an acceptable interim solution.

I think using the make rules instead of a direct `docker build` / `docker push` might be slightly preferable in terms of consistency. Maybe debatable.

test-resources/kind-config.yaml
```diff
@@ -1,6 +1,8 @@
-# See CONTRIBUTING.md -> "Semi-Manual Test Scenarios" and "Running End-to-End Tests on kind" for instructions to run
-# semi-manual and e2e tests on a kind cluster.
-
+# See CONTRIBUTING.md -> "Setting Up a Kind Cluster for Local Testing" for instructions.
```
basti1302: This is actually the file previously known as test-resources/kind-config-lr.yaml.

ten15bit (approved):
👍

PR #617, opened 2025-11-20
fix(collector): do not fail filelog-offset-volume-ownership on AWS EFS

basti1302:
Fix the init container filelog-offset-volume-ownership crash on AWS EFS based volumes.

In particular, replace the inline container command with an entrypoint script that deliberately runs without set -e, that is, it will not crash the init container if the chown command is not successful.

See images/filelogoffsetvolumeownership/README.md for a comprehensive discussion on the constraints and resulting design choices.

### DIFF
--- a/images/filelogoffsetvolumeownership/entrypoint.sh
+++ b/images/filelogoffsetvolumeownership/entrypoint.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+# SPDX-FileCopyrightText: Copyright 2025 Dash0 Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+# See README.md for an explanation of this script and its purpose.
+
+# MAINTENANCE NOTE: THIS SCRIPT *DELIBERATELY* DOES NOT RUN WITH set -e! See README.md for more details.
+set -x
+
+/bin/mkdir -p /var/otelcol/filelogreceiver_offsets
+/bin/chown -R 65532:0 /var/otelcol/filelogreceiver_offsets
+
+# log directory ownership and permission information for troubleshooting purposes
+ls -lad /var/otelcol/filelogreceiver_offsets
+ls -la /var/otelcol/filelogreceiver_offsets/
--- a/internal/collectors/otelcolresources/desired_state.go
+++ b/internal/collectors/otelcolresources/desired_state.go
@@ -1113,14 +1113,10 @@ func assembleFileLogVolumeOwnershipInitContainer(
 	config *oTelColConfig,
 	filelogOffsetsVolume corev1.Volume,
 ) corev1.Container {
+	// See images/filelogoffsetvolumeownership/README.md for information on this init container.
 	initFilelogOffsetVolumeOwnershipContainer := corev1.Container{
 		Name:  "filelog-offset-volume-ownership",
 		Image: config.Images.FilelogOffsetVolumeOwnershipImage,
-		Command: []string{
-			"/bin/sh",
-			"-c",
-			fmt.Sprintf("/bin/mkdir -p %s && /bin/chown -R %d:%d %s", offsetsDirPath, defaultUser, defaultGroup, offsetsDirPath),
-		},
 		Env: []corev1.EnvVar{
 			k8sNodeNameEnvVar,
 		},
--- a/images/filelogoffsetvolumeownership/Dockerfile
+++ b/images/filelogoffsetvolumeownership/Dockerfile
@@ -1,11 +1,17 @@
 # SPDX-FileCopyrightText: Copyright 2025 Dash0 Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-# Note: This basically just re-tags the busybox. We do this for consistency with other images. That is, all images used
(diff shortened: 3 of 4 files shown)


images/filelogoffsetvolumeownership/entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+# SPDX-FileCopyrightText: Copyright 2025 Dash0 Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+# See README.md for an explanation of this script and its purpose.
+
+# MAINTENANCE NOTE: THIS SCRIPT *DELIBERATELY* DOES NOT RUN WITH set -e! See README.md for more details.
+set -x
+
+/bin/mkdir -p /var/otelcol/filelogreceiver_offsets
+/bin/chown -R 65532:0 /var/otelcol/filelogreceiver_offsets
```
basti1302: This was previously in an inline command in internal/collectors/otelcolresources/desired_state.go.

internal/collectors/otelcolresources/desired_state.go
```diff
@@ -1113,14 +1113,10 @@ func assembleFileLogVolumeOwnershipInitContainer(
...
 ) corev1.Container {
+	// See images/filelogoffsetvolumeownership/README.md for information on this init container.
 	initFilelogOffsetVolumeOwnershipContainer := corev1.Container{
 		Name:  "filelog-offset-volume-ownership",
 		Image: config.Images.FilelogOffsetVolumeOwnershipImage,
-		Command: []string{
-			"/bin/sh",
-			"-c",
-			fmt.Sprintf("/bin/mkdir -p %s && /bin/chown -R %d:%d %s", offsetsDirPath, defaultUser, defaultGroup, offsetsDirPath),
-		},
```
basti1302: This is now in images/filelogoffsetvolumeownership/entrypoint.sh

test-resources/bin/lib/util
```diff
@@ -388,13 +388,27 @@ push_all_images() {
 }
 
 deploy_filelog_offsets_pvc() {
-  echo "STEP $step_counter: install filelog offset persistent volume claim"
-  if [[ "$is_kind_cluster" = "true" ]]; then
```
basti1302: We actually had one last instance of `$is_kind_cluster` here, which we missed in commit 35b9d1376f4e508ebf424c6b58c064db10603fe8. It would crash the script due to the unbound variable.

Since FILELOG_OFFSETS_PVC supports different values (including `kind`, `docker`, etc.) now instead of only true vs. false, this was rather easy to resolve.

test-resources/kind-config-lr.yaml
```diff
@@ -28,6 +28,8 @@ containerdConfigPatches:
 nodes:
   - role: control-plane
     extraMounts:
+      # Note: This volume is required on every node for testing filelog offset storage volumes, e.g.
+      # FILELOG_OFFSETS_PVC=kind test-resources/bin/test-scenario-xxx.sh
```
basti1302: @ ten15bit: I think you asked about this the other day. Without the extraMounts, when running a test script with `FILELOG_OFFSETS_PVC=kind`, things will not work properly.

That is, this:

```
IMAGE_REPOSITORY_PREFIX="localhost:5001/" \
  TEST_IMAGE_REPOSITORY_PREFIX="localhost:5001/" \
  PUSH_BUILT_IMAGES=true \
  FILELOG_OFFSETS_PVC=kind test-resources/bin/test-scenario-02-operator-cr-aum.sh
```

... will lead to the daemonset collector pods getting stuck in `Init:CreateContainerConfigError`:

```
Events:                                                                                                                                                                                                          Type     Reason            Age                From               Message                                                                                                                                       Warning  FailedScheduling  71s                default-scheduler  0/3 nodes are available: pod has unbound immediate PersistentVolumeClaims. not found
...
Warning  Failed            5s (x6 over 60s)   kubelet            Error: stat /offset-storage/: no such file or directory
```

This is due to the volume that gets added to the pods:
```
volumes:
- name: filelogreceiver-offsets
  persistentVolumeClaim:
    claimName: offset-storage-claim
```

The PVC & PV are defined in test-resources/filelog-offset-storage/offset-storage-pvc-kind.yaml.

images/filelogoffsetvolumeownership/entrypoint.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+# SPDX-FileCopyrightText: Copyright 2025 Dash0 Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+# See README.md for an explanation of this script and its purpose.
+
+# MAINTENANCE NOTE: THIS SCRIPT *DELIBERATELY* DOES NOT RUN WITH set -e! See README.md for more details.
+set -x
```
basti1302: This script and the omission of `set -e` here is the central change of this PR. IMO, under ordinary circumstances, omitting `set -e` is a non-starter, but in this specific case it might be the correct choice.

As the comment says, see the README.md next to this script an in-depth explanation.

PR #633, opened 2025-11-27
docs: add some clarifications to the docs

ten15bit:
- clearer description of what is currently supported wrt Prometheus CRDs
- adds comments for `error_mode` based on the description that's currently only in the code/crd and therefore not very visible


basti1302 (approved):
👍

PR #649, opened 2025-12-02
test(e2e): fix collection of kubectl output for test failures

basti1302:
This is a follow up to 9fe289eaf864494cb1a7f85eda2146ad3a0577d2. Using Gingko's ContinueOnFailure was not compatible with the existing approach to collect diagnostic information in case of failing test cases (RegisterFailHandler with a custom handler). The new approach (JustAfterEach hook) is more idiomatic anyway.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -54,6 +54,9 @@ var _ = Describe("Dash0 Operator", Ordered, ContinueOnFailure, func() {
 		} else {
 			e2ePrint("%s not found, assuming required environment variables are set via other means\n", dotEnvFile)
 		}
+
+		createDirAndDeleteOldCollectedLogs()
+
 		e2eKubernetesContext = os.Getenv("E2E_KUBECTX")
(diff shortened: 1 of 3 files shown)


test/e2e/e2e_test.go
```diff
@@ -118,9 +121,11 @@ var _ = Describe("Dash0 Operator", Ordered, ContinueOnFailure, func() {
...
 	})
 
-	BeforeEach(func() {
-		By("running BeforeEach hook of test suite root")
-		createDirAndDeleteOldCollectedLogs()
+	JustAfterEach(func() {
+		currentSpecReport := CurrentSpecReport()
+		if currentSpecReport.Failed() {
+			collectPodInfoAndLogs(currentSpecReport)
+		}
```
basti1302: This is the replacement for RegisterFailHandler with a custom handler. See https://onsi.github.io/ginkgo/#separating-diagnostics-collection-and-teardown-justaftereach. It's actually meant exactly for what we are doing here.

PR #655, opened 2025-12-04
feat(collector): enable pprof extension on demand

basti1302:
This adds a new option to the Helm values to enable the pprof extension both in the daemonset collector pods as well as in the collector deployment. This can be enabled temporarily to troubleshoot the collector's memory consumption.


ten15bit (approved):
👍

PR #661, opened 2025-12-09
feat(collectors): option to disable k8sattributes' replicaset informer

basti1302:
The k8sattributes processor makes the collector consumes a lot of memory at startup in clusters with an excessive amount of replicasets. This is due to how the processor uses a replicaset informer to derive k8s.deployment.name and k8s.deployment.uid for pods. Even in the daemonset collector, with filter.node_from_env_var being set, it needs to fetch all replicasets in the cluster.

This commit adds an option to the Helm chart to disable the replicaset informer within the k8sattributes processor. As a consequence, k8s.deployment.uid is then also removed from the list k8sattributes.extract.metadata, as this is required to actually make sure the costly replicaset informer is not used.

See here for an extensive discussion around the k8sattributes processor and the replicaset informer specifically:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/23067
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44407
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44708
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/36234
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42534

For now, this is only an option in the Helm chart. We can later add it as a Dash0OperatorConfiguration property, if needed. This depends on whether (and how fast) currently ongoing attempts to fix this issue upstream are successful.

Also: Get rid of one k8sattributes processor instance by moving it from the logs/otlp and the logs/filelog pipeline to the logs/downstream pipeline.

### DIFF
--- a/helm-chart/dash0-operator/templates/operator/deployment-and-webhooks.yaml
+++ b/helm-chart/dash0-operator/templates/operator/deployment-and-webhooks.yaml
@@ -233,6 +233,10 @@ spec:
         - name: OTEL_COLLECTOR_SEND_BATCH_MAX_SIZE
           value: {{ .Values.operator.collectors.sendBatchMaxSize | quote }}
         {{- end }}
+        {{- if .Values.operator.collectors.disableReplicasetInformer }}
+        - name: OTEL_COLLECTOR_K8SATTRIBUTES_DISABLE_REPLICASET_INFORMER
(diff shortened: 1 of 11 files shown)


bripkens (approved):
LGTM 👍

PR #664, opened 2025-12-11
test: add script to create a large number of Kubernetes entities

basti1302:
This can be used to test the resource consumption of the operator (and the collectors it manages) in relation to the number of Kubernetes entities and their rate of change.


basti1302:
I guess, strictly speaking, we do not need to have this in the repo. Let me know whether you think we should add it, otherwise I can also put this somewhere else.

PR #666, opened 2025-12-12
feat(target-allocator): add support for mTLS

ten15bit:
To support scraping endpoins that require authorization, the communication between the target-allocator and the collectors needs to be secured with mTLS so the credentials can be transferred safely.

### DIFF
--- a/helm-chart/dash0-operator/README.md
+++ b/helm-chart/dash0-operator/README.md
@@ -148,7 +148,7 @@ The operator supports the following CRDs
 #### Authorization
 
 If the scraped endpoints require authorization, it is mandatory to configure mTLS for the communication between the
-OpenTelemetry TargetAllocator and the collectors.
+OpenTelemetry Target Allocator and the collectors, so the credentials can be transfered in a secure manner.
 
 We recommend to use [cert-manager](https://cert-manager.io/) for the creation of the certificates/secrets, but any
 secrets following the [kubernetes.io/tls](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets) secret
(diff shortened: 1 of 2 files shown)


helm-chart/dash0-operator/README.md
```diff
@@ -145,7 +145,32 @@ The operator supports the following CRDs
 - `PodMonitor`
 - `ScrapeConfig` with `kubernetesSDConfigs`
 
-The scraping of endpoints requiring authorization is currently not supported, but will be added in an upcoming release.
+#### Authorization
+
+If the scraped endpoints require authorization, it is mandatory to configure mTLS for the communication between the
+OpenTelemetry TargetAllocator and the collectors.
```
basti1302: _question:_ Would it make sense to also explicitly mention the reason for this requirement for mTLS? (I.e. that the target allocator provides the auth credentials for the scraping from the CRs to the collectors in its HTTP response.) It is mentioned briefly in values.yaml, but I think it might be good to add a note about this here as well.

internal/targetallocator/taresources/desired_state.go
```diff
@@ -274,8 +289,25 @@ func assembleClusterRoleBinding(c *targetAllocatorConfig) *rbacv1.ClusterRoleBin
...
+		},
+	}
+
+	if extraConfig.TargetAllocatorMtlsEnabled {
+		ports = append(ports, corev1.ServicePort{
+			Name:       "https",
+			Protocol:   corev1.ProtocolTCP,
+			Port:       443,
+			TargetPort: intstr.FromString("https")})
+	}
```
basti1302: _question:_ If mTLS is enabled, will only port 443 be used? Should we still list port 80 for non-TLS traffic additionally in that setup, or should we rather only have port 443?

In particular, would the scrape configs including the auth information be made available via port 80 right now and would that allow bypassing mTLS?

helm-chart/dash0-operator/README.md
ten15bit: added a short note

internal/targetallocator/taresources/desired_state.go
basti1302: Oh, that's smart 👍

PR #669, opened 2025-12-15
chore: maintain our own target-allocator image

ten15bit:


basti1302:
I think we might want to add the new image also to these two places:
- https://github.com/dash0hq/dash0-operator/blob/0.91.0/test-resources/bin/lib/util#L142-L165 and below, plus https://github.com/dash0hq/dash0-operator/blob/0.91.0/test-resources/bin/lib/util#L539-L550 -- currently the test scenarios would not add parameters for the target allocator image when deploying the operator.
- https://github.com/dash0hq/dash0-operator/blob/0.91.0/test/e2e/container_images.go#L45-L76 / https://github.com/dash0hq/dash0-operator/blob/0.91.0/test/e2e/container_images.go#L97-L144 -- currently the end to end tests would use `ghcr.io/dash0hq/target-allocator:0.0.0`, even when `IMAGE_REPOSITORY_PREFIX="" is set.

PR #678, opened 2025-12-18
feat(dashboards): send the full resource including metadata

basti1302:


internal/controller/perses_dashboards_controller.go
```diff
@@ -451,46 +451,19 @@ func (r *PersesDashboardReconciler) MapResourceToHttpRequests(
...
 		}
+		r.setDisplayNameIfMissing(preconditionChecksResult, display)
 
-		// Remove all unnecessary metadata (labels & annotations), we basically only need the dashboard spec.
-		serializedDashboard, _ := json.Marshal(
-			map[string]interface{}{
-				"kind": "PersesDashboard",
-				"spec": specOrConfig,
-			})
+		serializedDashboard, _ := json.Marshal(dashboard)
```
basti1302: This change is the crux of this PR, instead of only sending `spec` we now send the full resource as the payload.

PR #679, opened 2025-12-18
fix(collectors): fix kubectl edit configmap for deployment collector

basti1302:
Also, add a make recipe for linting this for simple cases like the one fixed in this commit.


basti1302:
This is how `kubectl edit configmap` (or the equivalent view in k9s) looks if the configmap contains the problematic pattern.

It's not pretty. 🙂

image

For comparison, this is how kubectl edit configmap is supposed to look like.

image

ten15bit (approved):

PR #690, opened 2025-12-24
feat(workloads): do not replace existing OTEL_EXPORTER_OTLP_* settings

basti1302:
Do not replace OTEL_EXPORTER_OTLP_ENDPOINT or
OTEL_EXPORTER_OTLP_PROTOCOL, when either of them is set, leave both settings unchanged when instrumenting containers.

If a subset of containers in a workload cannot be instrumented
successfully, this is now documented with the Kubernetes event which is
attached to the workload.


basti1302:
@claude review

PR #701, opened 2026-01-07
fix(views): use simple string type for filter.value and filter.values

basti1302:
Also for implicitFilter.value/implicitFilter.values.

The Dash0 API's OpenAPI spec allows either simple strings or AnyValue for filter values via oneOf, but Kubernetes CRDs do not support oneOf.

This changes the type for the four mentioned attributes from AnyValue to plain strings. This works for the core use case around this, that is, downloading a view as yaml from the Dash0 UI and then putting it into a Kubernetes cluster as a Dash0View resource. (The download functionality should only generate the string variant, and never the AnyValue variant.)


ten15bit (approved):
👍

PR #703, opened 2026-01-08
fix(checkrules): do not re-use origins of unrelated check rules

basti1302:
We previously used the group name and the index of the rule within the group to compose the check rule origin. When individual rules are deleted from a group, or rules are moved around within a group, check rules might be assigned an origin that previously belonged to a different rule. This is problematic, because check rules have a history of check executions (and a history of failed checks) in Dash0. Rule A re-using the origin of an unrelated rule B would lead to rule A displaying the history of rule B in the UI.

By using the rule's alert name in the origin, this can no longer happen. The index of the rule within the group is no longer used. Duplicate origins due to duplicate alert names within a group are avoided by appending a counter to rules with duplicated alert names.

This has a couple of consequences:
- rules no longer take over the origin of other unrelated rules
- renaming a rule will discard the history of checks

BREAKING CHANGE: This is a breaking change in so far as the origin of _every_ check rule in the system will change once, when the operator is updated to a release that contains this change. All failed check history will be effectively discarded. This is a one time event, which cannot be avoided when fixing the issue outlined above.

### DIFF
--- a/internal/controller/api_sync_common.go
+++ b/internal/controller/api_sync_common.go
@@ -39,39 +39,46 @@ type ResourceToRequestsResult struct {
 	// the total number of eligible items in the Kubernetes resource
 	ItemsTotal int
 	// the request objects for which the conversion was successful
-	HttpRequests []HttpRequestWithItemName
+	ApiRequests []WrappedApiRequest
 	// only set for resource types with a one-to-many relationship between Kubernetes resources and Dash0 entities,
 	// contains a list of all Dash0 origins produced by this Kubernetes resource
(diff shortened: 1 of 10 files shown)


basti1302:
Note: There are two commits in this PR, which are only tangentially related. It might be easier to review the commits individually. I can also split them up into two separate PRs if needed.

basti1302:
This is planned for a separate release after rolling out https://github.com/dash0hq/dash0-operator/pull/707.

ten15bit (approved):
👍

PR #707, opened 2026-01-11
feat: add support for namespaced exporters

ten15bit:
## Overview

Adds support for per-namespace export settings via the `Dash0Monitoring` resource.

The export settings from the `Dash0OperatorConfiguration` serve as defaults that can be overridden by providing an `export` in a namespace's `Dash0Monitoring`.

## How to test

Besides running the unit and e2e tests, the quickest way to test it is to create a dataset and a token with ingestion permissions for that dataset and set the env vars `DASH0_NAMESPACED_DATASET` and `DASH0_NAMESPACED_AUTHORIZATION_TOKEN` in addition to the common env vars used by the test scripts.

After that, running

```sh
test-resources/bin/test-scenario-07-namespaced-exporter.sh
```

will install the operator with a config resource using the default export config and a monitoring resource in a test namespace that uses the custom settings.

In Dash0, you should see data in both the default dataset (coming from the operator itself and cluster telemetry) and in the namespaced dataset coming from the monitored workload in that namespace.

### DIFF
--- a/internal/collectors/otelcolresources/authorization.go
+++ b/internal/collectors/otelcolresources/authorization.go
@@ -1,0 +1,71 @@
+// SPDX-FileCopyrightText: Copyright 2026 Dash0 Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package otelcolresources
+
+import (
+	"fmt"
+	"strings"
+
+	dash0common "github.com/dash0hq/dash0-operator/api/operator/common"
+	dash0v1alpha1 "github.com/dash0hq/dash0-operator/api/operator/v1alpha1"
+	dash0v1beta1 "github.com/dash0hq/dash0-operator/api/operator/v1beta1"
+)
+
+const authEnvVarPrefix = "OTELCOL_AUTH_TOKEN"
+
+var authEnvVarNameDefault = fmt.Sprintf("%s_DEFAULT", authEnvVarPrefix)
+
+type dash0ExporterAuthorization struct {
+	EnvVarName    string
+	Authorization dash0common.Authorization
+}
+
+type dash0ExporterAuthorizationByNamespace = map[string]dash0ExporterAuthorization
+
+type dash0ExporterAuthorizations struct {
+	DefaultDash0ExporterAuthorization     *dash0ExporterAuthorization
+	NamespacedDash0ExporterAuthorizations dash0ExporterAuthorizationByNamespace
+}
+
+func (auths dash0ExporterAuthorizations) all() []dash0ExporterAuthorization {
+	allAuths := make([]dash0ExporterAuthorization, 0)
+	if auths.DefaultDash0ExporterAuthorization != nil {
+		allAuths = append(allAuths, *auths.DefaultDash0ExporterAuthorization)
+	}
+	for _, auth := range auths.NamespacedDash0ExporterAuthorizations {
+		allAuths = append(allAuths, auth)
+	}
+	return allAuths
+}
+
+func authEnvVarNameForNs(namespace string) string {
+	envVarSuffix := strings.ReplaceAll(namespace, "-", "_")
+	envVarSuffix = strings.ToUpper(envVarSuffix)
+
+	return fmt.Sprintf("%s_NS_%s", authEnvVarPrefix, envVarSuffix)
+}
+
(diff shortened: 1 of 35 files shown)


internal/collectors/otelcolresources/daemonset.config.yaml.template
```diff
@@ -29,11 +32,55 @@ extensions:
...
+  forward/traces-processors: {}
+  forward/metrics-processors: {}
+  forward/logs-processors: {}
+  forward/traces-default-exporter: {}
+  forward/metrics-default-exporter: {}
+  forward/logs-default-exporter: {}
+
+  {{- $hasNamespacedExporters := gt (len .NamespacedExporters) 0 }}
+
+  {{- if $hasNamespacedExporters }}
```
ten15bit: unfortunately, a routing connector with only a default pipeline is not valid

so when there are no namespaced exporters we need to skip the entire block and use the `forward/*-default-exporter` connector instead

internal/collectors/otelcolresources/exporter.go
```diff
@@ -0,0 +1,194 @@
...
+	if export == nil {
+		return nil, nil
+	}
+	if !isDefault && namespace == nil {
+		return nil, fmt.Errorf("no valid nameSuffix provided for namespaced exporter, unable to create OpenTelemetry collector")
+	}
+
+	var nameSuffix string
+	if isDefault {
+		nameSuffix = "default"
```
basti1302: _question:_ Could this suffix collide with a namespaced exporter in the `default` namespace? (Could totally also be me reading the code wrong, just double checking...)

internal/collectors/otelcolresources/deployment.config.yaml.template
```diff
@@ -370,7 +370,7 @@ service:
       {{- if (or .DevelopmentMode .DebugVerbosityDetailed) }}
       - debug
       {{- end }}
-      {{- range $i, $exporter := .Exporters }}
+      {{- range $i, $exporter := .DefaultExporters }}
```
basti1302: _question:_ There is no routing for the telemetry produced by the deployment. Why is that? I think some of the telemetry produced by the collector deployment is also namespaced. Wouldn't we want to route it in the same way we do in the DaemonSet?

test/e2e/e2e_test.go
```diff
@@ -725,6 +722,230 @@ var _ = Describe("Dash0 Operator", Ordered, ContinueOnFailure, func() {
...
+						false,
+						timestampLowerBound,
+						"/ready",
+						"", // health check spans have no query parameter
+						"",
+					)
+				})
+			})
+
+			Describe("telemetry transformation", func() {
```
basti1302: same here, this should be in "without a deployed Dash0 monitoring resource"

internal/collectors/otelcolresources/exporter.go
basti1302: Right, my bad, makes sense 👍

internal/collectors/otelcolresources/authorization.go
```diff
@@ -0,0 +1,71 @@
...
+	envVarSuffix := strings.ReplaceAll(namespace, "-", "_")
+	envVarSuffix = strings.ToUpper(envVarSuffix)
+
+	return fmt.Sprintf("%s_NS_%s", authEnvVarPrefix, envVarSuffix)
+}
+
+func collectDash0ExporterAuthorizations(operatorConfigurationResource *dash0v1alpha1.Dash0OperatorConfiguration,
+	allMonitoringResources []dash0v1beta1.Dash0Monitoring) dash0ExporterAuthorizations {
+	var defaultAuth *dash0ExporterAuthorization = nil
+	if operatorConfigurationResource.Spec.Export.Dash0 != nil {
```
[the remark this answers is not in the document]
ten15bit: I will add an additional nil check, but as you said, we validate the config resource very early and won't continue if there is no export.

I think it would be nicer to validate things once and use types to ensure they are valid in later stages (instead of rechecking again).

internal/collectors/otelcolresources/deployment.config.yaml.template
basti1302: That is a fair point. This is actually a much more tricky question than I anticipated.

basti1302 (approved):
Great work! 💯

PR #721, opened 2026-01-17
fix(instrumentation): clean up LD_PRELOAD entry for Dash0 injector

basti1302:
This is a follow up to b8fe550b9be30667a9d06dfa35bdf54a6367448d.

Workloads instrumented by operator versions prior to 0.97.0 might still reference `/__dash0__/dash0_injector.so` in `LD_PRELOAD`, in addition to the correct entry (`/__otel_auto_instrumentation/injector/libotelinject.so`). While this is harmless (non-existing files in LD_PRELOAD are ignored by the linker), the operator now removes this obsolete entry.


ten15bit (approved):
👍

PR #724, opened 2026-01-19
test(e2e): split e2e test off into its own go module

basti1302:
Avoid listing e2e test dependencies and replace directives in the operator manager's go.mod file.


ten15bit (approved):
nice 🎉

PR #739, opened 2026-01-27
feat(instrumentation): Python auto-instrumentation support (opt-in)

basti1302:


images/instrumentation/python/usercustomize.py
```diff
@@ -0,0 +1,79 @@
...
+from os.path import dirname
+from sys import path, version, version_info
+
+def import_distro():
+    print("[usercustomize] import_distro")
+    # print("[usercustomize] PYTHONPATH: {}".format(os.environ['PYTHONPATH']))
+    current_site = dirname(__file__)
+    print("[usercustomize] current_site: {}".format(current_site))
+    print("[usercustomize] version_info: {}".format(version_info))
+    print("[usercustomize] version_info: {}".format(version_info))
```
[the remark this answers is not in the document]
basti1302: Thanks for catching this, this is actually excessive debug info that shouldn't have been committed in the first place. I'll clean it up.

basti1302:

images/instrumentation/python/usercustomize.py
```diff
@@ -0,0 +1,153 @@
...
+
+def _print_debug_msg(message):
+    if debug_enabled:
+        _print_to_stderr(message)
+
+_print_debug_msg("running usercustomize.py")
+
+def _print_cannot_auto_instrument_message(reason):
+    if hasattr(sys, 'argv'):
+        _print_to_stderr("warning: cannot auto-instrument Python process: {} [{}]".format(reason, ' '.join(sys.argv)))
```
mmanciop: This line deserves a comment IMO

images/instrumentation/python/usercustomize.py
```diff
@@ -0,0 +1,153 @@
...
+    else:
+        _print_to_stderr("warning: cannot auto-instrument Python process: {}".format(reason))
+
+def _read_requirements_txt():
```
mmanciop: `requirements.txt` does not list transitive dependencies and there is stuff in there we really should be careful about beyond the `google` and `protobuf` stuff, like `prometheus_client`. From the top of my head, I could imagine a `Pipfile.lock`, which is JSON, should be easy to deal with.

images/instrumentation/test/python-2.7/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# SPDX-FileCopyrightText: Copyright 2026 Dash0 Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+ARG instrumentation_image=dash0-instrumentation:latest
+ARG base_image=python:3.13.7-slim
```
mmanciop: This Dockerfile deserves a different default :-)

images/instrumentation/python/usercustomize.py
mmanciop: Nevermind, I had missed the recursive analysis later on. I might be less confusing with listing also transitive dependencies explicitly, but nevermind.

images/instrumentation/python/usercustomize.py
```diff
@@ -0,0 +1,153 @@
...
+
+    # We cannot use `sys.version_info.major` or other named attributes, as they only got introduced only in Python 3.1.
+    if version_info[0] != 3 or version_info[1] < 9:
+        path.remove(current_site)
+        _print_cannot_auto_instrument_message("unsupported Python version: {}".format(version))
+        return
+    _print_debug_msg("found eligible Python version: {}".format(version_info))
+
+    otlp_protocol = os.environ.get('OTEL_EXPORTER_OTLP_PROTOCOL')
+    if otlp_protocol is None:
```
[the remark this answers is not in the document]
basti1302: I'll add something there. But just to be clear, this cannot happen if the workload is entirely instrumented by the operator, since it will set exactly the OTEL_EXPORTER_OTLP_PROTOCOL value we are checking for here.

It can only happen, if the workload has a pre-existing OTEL_EXPORTER_OTLP_PROTOCOL or OTEL_EXPORTER_OTLP_ENDPOINT setting.

images/instrumentation/Dockerfile
```diff
@@ -55,17 +55,23 @@ FROM python:3.14.2-trixie AS build-python-glibc
...
 FROM python:3.14.2-alpine3.23 AS build-python-musl
 RUN mkdir -p /dash0-init-container/python
 WORKDIR /dash0-init-container/python
 COPY python/requirements.txt .
+COPY python/extract-flattened-requirements.sh .
 ENV PIP_DISABLE_PIP_VERSION_CHECK=1
-RUN pip install --quiet --root-user-action ignore --target . -r requirements.txt
+RUN pip install --quiet --root-user-action ignore --target . -r requirements.txt && \
+  ./extract-flattened-requirements.sh && \
+  rm extract-flattened-requirements.sh
```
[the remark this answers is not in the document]
basti1302: I wonder if we should merge this with the non-transitive dependencies that come directly from `requirements.txt`. Currently we would only check against indirect dependencies.

PR #758, opened 2026-01-31
ci: run e2e tests on kind

ten15bit:
This PR adds a simple workflow to run the e2e tests on kind as part of the CI.

Future improvements: caching (which would also be relevant for the main CI workflow) or using the already built images from ghcr (when running on main)

successful run: https://github.com/dash0hq/dash0-operator/actions/runs/21862087509/job/63093616044


ten15bit:
needs #759 to be compatible with kind >= 0.31.0 which is installed on the runner

PR #767, opened 2026-02-04
feat: collect namespace labels and annotations as resource attributes

basti1302:
Add opt-in to collect namespace labels and annotations as additional resource attributes.


ten15bit (approved):
👍

PR #803, opened 2026-02-17
feat: support multiple exports for telemetry and api sync

ten15bit:
## Overview

This PR adds support for using multiple exports for telemetry and API sync.

Example:

```yaml
apiVersion: operator.dash0.com/v1alpha1
kind: Dash0OperatorConfiguration
metadata:
  name: dash0-operator-configuration-resource
spec:
  collectNamespaceLabelsAndAnnotations:
    enabled: false
  collectPodLabelsAndAnnotations:
    enabled: true
  exports:
    - dash0:
        apiEndpoint: https://api.europe-west4.gcp.dash0-dev.com
        authorization:
          secretRef:
            key: token
            name: dash0-authorization-secret
        dataset: default
        endpoint: ingress.europe-west4.gcp.dash0-dev.com:4317
    - dash0:
        apiEndpoint: https://api.europe-west4.gcp.dash0-dev.com
        authorization:
          secretRef:
            key: token
            name: dash0-authorization-secret-for-second-dataset
        dataset: second-dataset
        endpoint: ingress.europe-west4.gcp.dash0-dev.com:4317
  ...
```

This would enable export and API sync to two different datasets at the same time. `gRPC` and `http` exports are also supported.

## How to test

Besides the automated tests, the functionality can be tested by e.g. creating a second dataset and token and setting

```
export DEPLOY_OPERATOR_CONFIGURATION_VIA_HELM="false"
export USE_MULTI_CAST="true"
export DASH0_SECOND_DATASET="<name_of_second_dataset>"
export DASH0_SECOND_AUTHORIZATION_TOKEN='<token_for_second_dataset>'
```

and then running one of the test scripts like `test-resources/bin/test-scenario-01-aum-operator-cr.sh`

This will deploy an operator config with two exports and telemetry, dashboards, etc... should be visible in both datasets. Afterwards modify the dash0 operator configuration and/or monitoring resource as needed to test different cases.

### DIFF
--- a/helm-chart/dash0-operator/README.md
+++ b/helm-chart/dash0-operator/README.md
@@ -663,8 +663,9 @@ spec:
 
 ### Namespace-specific Exports and API-sync Overrides
 
-It is possible to override the global/default exports config provided via the Dash0 operator configuration on a per-namespace
-basis by providing the corresponding exports config via the Dash0 monitoring resource.
+It is possible to override the global/default exports config provided via the Dash0 operator configuration on a
+per-namespace basis by providing the corresponding exports config via the Dash0 monitoring resource. Please note that
+the namespace-specific overrides replace the default list of exports from the Dash0 operator configuration.
 
 The override supports both the export of telemetry and the sync of resources, like dashboards or views, via the Dash0 API.
 
@@ -1405,12 +1406,9 @@ spec:
             value: my-value
 ```
 
-You can combine up to three exporters (i.e. Dash0 plus gRPC plus HTTP) to send data to multiple backends. This allows
-sending the same data to two or three targets simultaneously. At least one exporter has to be defined. More than three
-exporters cannot be defined. Listing two or more exporters of the same type (i.e. providing `spec.export.grpc` twice)
-is not supported.
-
-Here is an example that combines three exporters:
+Export to multiple backends is also supported. The supplied backends can be either of the same type or of different
+types. In the following example the telemetry would be sent to two different datasets in Dash0 and in addition to a gRPC
+endpoint:
 
 ```yaml
 apiVersion: operator.dash0.com/v1alpha1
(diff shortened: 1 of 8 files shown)


internal/collectors/otelcolresources/collector_config_maps_test.go
```diff
@@ -627,6 +728,244 @@ var _ = Describe("The OpenTelemetry Collector ConfigMaps", func() {
 			Expect(dash0OtlpExporter["endpoint"]).To(Equal(EndpointDash0TestAlternative))
 		}, daemonSetAndDeployment)
 
+		DescribeTable("should render two Dash0 default exporters from multiple Exports entries", func(cmTypeDef configMapTypeDefinition) {
```
basti1302: _question:_ With the changes from the most recent commit https://github.com/dash0hq/dash0-operator/pull/803/changes/7014a9a62ab4de31a9fe5a32b4ad0772d40e22db, would it make sense to add tests for the config map that also explicitly test for a stable order of exports?

We currently read the exporter from the result of `parseConfigMapContent`, at which point the exporters are in a map and a verification of the order is not possible based on the map.

But we could add checks whether exporter A appears before exporter B in the raw string `configMap.Data[dataKey]`.

OTOH, maybe this is overkill? WDYT?

helm-chart/dash0-operator/README.md
```diff
@@ -659,10 +661,10 @@ spec:
   synchronizePrometheusRules: false
 ```
 
-### Namespace-specific Export and API-sync Overrides
+### Namespace-specific Exports and API-sync Overrides
 
-It is possible to override the global/default export config provided via the Dash0 operator configuration on a per-namespace
-basis by providing the corresponding export config via the Dash0 monitoring resource.
+It is possible to override the global/default exports config provided via the Dash0 operator configuration on a per-namespace
+basis by providing the corresponding exports config via the Dash0 monitoring resource.
```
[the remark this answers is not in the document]
ten15bit: ok, for me that's what `override` means, but I'll add a note :+1:

internal/collectors/otelcolresources/collector_config_maps_test.go
ten15bit: as discussed, I added it to the existing `"rendered objects must be stable"` test in`desired_state_test.go` that does a `deepEqual` of the entire state already

PR #810, opened 2026-02-19
fix(python): enable/disable Python auto-instrumentation immediately

basti1302:
When operator.instrumentation.enablePythonAutoInstrumentation is changed via helm upgrade, apply the new setting immediately to workloads in namespaces that have instrumentWorkloads.mode=all.


basti1302:
We might want to merge this after releasing https://github.com/dash0hq/dash0-operator/pull/803, I think.

PR #819, opened 2026-02-22
feat: target-allocator self-monitoring

ten15bit:
when `selfMonitoring` and `prometheusCrdSupport` are enabled, the following is added to the daemonset collector configs:

- an additional `filelogreceiver` for collecting target-allocator logs
- prometheus scrapeconfig to discover and scrape the target-allocator's metrics endpoint
- ta-specific exclusions, so the logs/metrics aren't dropped, even when there isn't a monitoring resource in the operator namespace


basti1302 (approved):
👍

PR #829, opened 2026-02-24
chore(collectors): replace deprecated k8sattributes with k8s_attributes, disable deprecated kubeletstats resource attributes 

basti1302:
This removes the warning:
"k8sattributes alias is deprecated; use k8s_attributes instead"


basti1302:
e2e test run: https://github.com/dash0hq/dash0-operator/actions/runs/22348082342

PR #830, opened 2026-02-24
do not merge: make lint takes 23 seconds locally, why does it take 4 minutes on CI??

basti1302:


basti1302:
Looking through https://github.com/golangci/golangci-lint-action/issues/297#issuecomment-2067601934, this is (a) more or less a known issue apparently and (b) we might want to switch to golangci/golangci-lint-action on CI instead of executing it via make. Closing this for now.

PR #851, opened 2026-03-06
feat: namespace auto-monitoring

basti1302:
Introduces a new feature to automatically monitor new namespaces. Previously, to monitor a namespace, you had to add a Dash0Monitoring resource to the namespace.

This feature allows to automatically monitor existing namespaces that do not have a monitoring resource yet, as well as new namespaces as they are created.

This feature is off by default, and has to be enabled explicitly.


internal/webhooks/monitoring_mutating_webhook.go
```diff
@@ -157,7 +158,9 @@ func (h *MonitoringMutatingWebhookHandler) setTelemetryCollectionRelatedDefaults
 	}
 
 	if monitoringSpec.InstrumentWorkloads.Mode == "" {
-		if telemetryCollectionEnabled {
+		if request.Namespace == h.operatorNamespace {
+			monitoringSpec.InstrumentWorkloads.Mode = dash0common.InstrumentWorkloadsModeNone
+		} else if telemetryCollectionEnabled {
```
basti1302: This is actually unrelated to the rest of the PR - if users add a monitoring resource to the operator namespace, we should not allow instrumenting workloads there.

api/operator/v1alpha1/operator_configuration_types.go
```diff
@@ -55,8 +57,8 @@ type Dash0OperatorConfigurationSpec struct {
 
 	// The configuration of the default observability backends to which telemetry data will be sent by the operator, as
 	// well as the backend that will receive the operator's self-monitoring data. This property is mandatory.
-	// These can either be Dash0 or another OTLP-compatible backends. Every Export entry can contain up to three distinct
-	// exporters (i.e. Dash0 plus gRPC plus HTTP).
+	// These can either be Dash0 or another OTLP-compatible backend. Every Export entry can contain up to three distinct
```
ten15bit: `This property is mandatory. These can either...` doesn't read very nicely I think.

maybe `Supported backends are Dash0 or any OTLP-compatible backend.`?

PR #871, opened 2026-03-17
fix: enabling api sync should trigger an initial sync of resources

ten15bit:
This PR fixes an issue in the following scenario:

- prometheus rules sync is disabled for a namespace
- a prometheus rule CR already exists
- sync for namespace is enabled
- one of the two scenarios happens
  - monitoring resource has no custom api config,  `SetNamespacedApiConfigs` is not called and sync is never triggered
  - because the api config itself is not changed, the check in `SetNamespacedApiConfigs` does not trigger a sync (this check is needed so we do not trigger a sync every time an unrelated change to the operator config or monitoring resource is made)
- as a result, the resource is not synced automatically (only after it's modified or the api config changes)

The same issue exists for Perses dashboards.

In order to fix that behavior, a new flag is added that keeps track of the current "sync enabled" state and if the state changes from `false` to `true` it ensures a sync is triggered regardless of whether the api config has changed too.

A small downside of the new behavior is that if a user changes the "sync enabled" state and the api config in one operation, it will do one unnecessary sync. However, the api config will in practice not be changed frequently (in most cases not at all), so that one additional sync should not cause any issues.

### DIFF
--- a/internal/controller/api_sync_common.go
+++ b/internal/controller/api_sync_common.go
@@ -61,6 +61,7 @@ type NamespacedApiClient interface {
 	SetNamespacedApiConfigs(context.Context, string, []ApiConfig, logd.Logger)
 	RemoveNamespacedApiConfigs(context.Context, string, logd.Logger)
 	SetSynchronizationEnabled(context.Context, string, *dash0v1beta1.Dash0Monitoring, logd.Logger)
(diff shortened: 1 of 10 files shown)


basti1302 (approved):
👍

PR #927, opened 2026-04-08
feat(chart): reduce RBAC permissions if telemetry collection is disabled

basti1302:
If telemetry collection is disabled via Helm chart values, only a reduced set of permissions will be set up by the operator Helm chart. RBAC permissions that are only required to manage collectors, the target allocator, or to auto-instrument workloads are not granted. This enables using the operator for IaC purposes (manage dashboards, check rules, synthetic checks and views via the operator) without granting the operator permissions that are not required for this use case.


p-d0 (commented):
I'm wondering whether it would be good to have a separate flag in the helm chart that gets passed to the operator at startup, so the operator is aware it's running in "sync-only" mode.

Then the validating webhooks could return a useful error if the user tries to enable something that won't work.

p-d0 (approved):

PR #931, opened 2026-04-09
refactor(iac): replace hand-rolled HTTP retry with dash0-api-client-go transport

mmanciop:
## Summary

- Replace bare `&http.Client{}` with `dash0.NewTransport(...).HTTPClient()` from `dash0-api-client-go` v1.8.0, configured with 2 retries (3 total attempts), 1s min / 3s max backoff
- Remove the operator's hand-rolled retry wrapper (`executeSingleHttpRequestWithRetryAndReadBody` + `RetryWithCustomBackoff`) in favor of the library's transport-level retry
- Remove `GetHttpRetryDelay()` from `ApiSyncReconciler` interface and all implementations
- Remove `httpRetryDelay` field and `overrideHttpRetryDelay` from all controllers

This is the first step toward rebasing the operator's IaC capabilities on top of `dash0-api-client-go`. Future PRs can add support for new CRDs using the library's higher-level typed APIs without touching the currently-supported CRDs.

## Bug fixes

This PR also fixes two latent bugs in the old retry logic:

1. **Request body not cloned on retry.** The old code retried the same `*http.Request` without resetting `req.Body`. After the first attempt, the body's `io.Reader` was already consumed, so retries of PUT requests (all upserts) would send an empty body. The library's retry transport properly clones the body via `req.GetBody()` before each retry.

2. **No context cancellation between retries.** The old code used `retry.OnError` which did not check for context cancellation between retry waits. If a reconciliation was cancelled (e.g. during shutdown), the operator would block until all retries completed. The library checks `req.Context().Done()` between attempts, allowing prompt cancellation.


basti1302 (changes requested):
Unit tests are currently failing. Plus, the removal of overrideHttpRetryDelay might be an issue for the unit tests. We used that to test the retry behavior in the unit tests and still have them run in a somewhat reasonable time. Having multiple 1-3 minute retries in the unit tests is not an option.

https://pkg.go.dev/github.com/dash0hq/dash0-api-client-go#readme-testing recommends [https://pkg.go.dev/github.com/dash0hq/dash0-api-client-go/dash0test ](https://pkg.go.dev/github.com/dash0hq/dash0-api-client-go/dash0testfor) for testing, but I haven’t looked into this in detail. In particular, I don't know whether this would allow to test the retry behavior without having 1-3 minutes of wait time in the unit tests.

basti1302 (approved):
LGTM now 👍

PR #939, opened 2026-04-13
chore(controller): spread out API sync requests

basti1302:
Rate-limit API sync requests on the client side to avoid hitting the backend's rate limit.

### DIFF
--- a/internal/controller/api_sync_common.go
+++ b/internal/controller/api_sync_common.go
@@ -13,13 +13,14 @@ import (
 	"slices"
 	"strings"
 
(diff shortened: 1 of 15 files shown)


basti1302:
It is somewhat questionable if we actually want to add this. We should either proceed with this PR or cancel OPE-248.

basti1302:
Closing since this has become stale and we basically decided to not go this way. (Instead, rate limits should be fixed backend side to avoid this as much as possible, plus we might want to introduce retries for synchronization errors/issues in general.)

PR #949, opened 2026-04-14
test: bump go.opentelemetry.io dependencies

basti1302:
Combines https://github.com/dash0hq/dash0-operator/pull/948 and https://github.com/dash0hq/dash0-operator/pull/946.


basti1302:
/run-e2e with an existing operator deployment and profiling enabled

PR #955, opened 2026-04-16
fix(self-monitoring): do not discard self-monitoring metrics in filter

basti1302:
The filter for discarding metrics for non-monitored namespaces would previously discard kubeletstat and k8s_receiver metrics from the operator manager and the collectors, even if self-monitoring was enabled. The filter now lets these metrics through.

Additionally, the condition to let target-allocator metrics through relied on pod labels. Collecting those can be disabled in the operator configuration resource. When disabled, the metrics for the target-allocator would have been discarded as well. This now relies on the deployment name resource attributes, which should always be set.

### DIFF
--- a/internal/collectors/otelcolresources/collector_config_maps.go
+++ b/internal/collectors/otelcolresources/collector_config_maps.go
@@ -225,8 +225,7 @@ func assembleCollectorConfigMap(
 	}
 	namespaceOttlFilter := renderOttlNamespaceFilter(
 		monitoredNamespaces,
-		config.SelfMonitoringConfiguration.SelfMonitoringEnabled,
-		config.PrometheusCrdSupportEnabled,
+		config,
 	)
 	customTelemetryFilters := aggregateCustomFilters(filters)
 	customTelemetryTransforms := aggregateCustomTransforms(transforms)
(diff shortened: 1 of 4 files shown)


internal/collectors/otelcolresources/collector_config_maps.go
```diff
@@ -317,20 +316,37 @@ func compressContent(collectorConfiguration string) (bytes.Buffer, error) {
...
 	// target-allocator metrics are considered self-monitoring and therefore controlled via the global SelfMonitoring
 	// config (and prometheusCrdSupportEnabled) from the Dash0OperatorConfiguration.
 	taExclusion := ""
-	if selfMonitoringEnabled && prometheusCrdSupportEnabled {
+	if selfMonitoringEnabled && config.PrometheusCrdSupportEnabled {
+		taDeploymentName := taresources.DeploymentName(config.TargetAllocatorNamePrefix)
 		taExclusion = fmt.Sprintf(
-			"(resource.attributes[\"k8s.pod.label.app.kubernetes.io/instance\"] != \"%s\" and "+
-				"resource.attributes[\"k8s.pod.label.app.kubernetes.io/name\"] != \"%s\") and ",
-			taresources.AppKubernetesIoInstanceValue, taresources.AppKubernetesIoNameValue)
```
basti1302, line 329: Collecting pod labels can be disabled, and then metrics for the target-allocator don't have these resource attributes, and they would be discarded by the filter. This now relies on the deployment name resource attributes, which should always be set. Just to double check, would you agree with that reasoning?

basti1302: Side note: I think kubeletstat and k8s_receiver metrics from the collector pods were previously available as a side effect of the taExclusion 🤔 

The way I read this is that metrics for any pod that matches app.kubernetes.io/instance=dash0-operator _OR_ app.kubernetes.io/name=opentelemetry-target-allocator would have gone through. And the collectors also have app.kubernetes.io/instance=dash0-operator. 

(The whole filter is a series of conditions concatenated by AND, the parentheses around the two pod label conditions were not doing anything, and the filter discards anything that matches the complete condition - or to put it differently, it lets everything pass through that matches the negation of at least one individual clause.)

p-d0 (approved):
Personally, I would prefer to not rely on the deployment name alone. In this case it will probably work since we use the helm release name as a prefix, but that's an implementation detail.

basti1302:
> Personally, I would prefer to not rely on the deployment name alone. In this case it will probably work since we use the helm release name as a prefix, but that's an implementation detail.

Fair point. Do you have an alternative suggestion?

p-d0:
> > Personally, I would prefer to not rely on the deployment name alone. In this case it will probably work since we use the helm release name as a prefix, but that's an implementation detail.
> 
> Fair point. Do you have an alternative suggestion?

We could add the operator ns in addition to the deployment name.

(or did I miss that somewhere? :thinking:)

basti1302:
> We could add the operator ns in addition to the deployment name.
> 
> (or did I miss that somewhere? 🤔)

Right, that's a good idea. 👍  Will do.

Somehow the [previous issue](https://github.com/dash0hq/dash0-operator/pull/955#discussion_r3093622527) made me want to go with a single condition for each workload we want to allow telemetry for, but this can be managed. (I think it just needs to be `or` inside the parentheses and then we should be good.)

basti1302:
@p-d0 I've added namespace conditions to the filter, please take another look.

basti1302:
I personally find the filter logic with nested conditions somewhat hard to reason about, so I also added a test that actually evaluates the OTTL filter against different telemetry to verify the filter works as intended.

PR #941, opened 2026-04-13
fix(chart): wait for AllowlistSynchronizer ready state

basti1302:
When running on Google GKE Autopilot, and the allowlist synchronizer resource is managed by the operator Helm chart, helm install/upgrade now waits for the allowlist synchronizer resource to be ready and to have the matching operator allowlist version.

Also:
- update all workloads to GKE allowlist version v1.0.3 and use "--dash0-" prefixes for most command line arguments.
- remove cloud.google.com/matching-allowlist annotation from post-delete-remove-allowlist-synchronizer job, as this can block removal of the terminating pod

PR #958, opened 2026-04-17
feat: introduce IE and add first set of IE components

p-d0:


test-resources/intelligentedge/operator-intelligent-edge-collector-builder-config.yaml
```diff
@@ -0,0 +1,78 @@
...
+    path: ./extension/dash0settingsonedgeextension
+
+receivers:
+  - gomod: go.opentelemetry.io/collector/receiver/otlpreceiver v0.147.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/filelogreceiver v0.147.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver v0.147.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/k8sclusterreceiver v0.147.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/k8seventsreceiver v0.147.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kubeletstatsreceiver v0.147.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver v0.147.0
```
p-d0, line 43: Upgrading to `0.150.0` didn't work because of incompatibilities with the prometheus fork (see `replaces` section) and we decided to skip `148` and `149`, so I kept it at `0.147.0` for now.

The prometheus and clickhouse forks can't be removed because `backend-utils` depends on them and that module is used by the IE components. Ideally `backend-utils` should be refactored into `backend` and `common` utils to be able to remove the dependency on the forks and make upgrading easier.

test-resources/intelligentedge/build-ie-and-barker.sh
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env bash
```
p-d0, line 1: This is just a temporary helper script, until we either build the images in the operator repo (imho that would be the preferred way), or build the images in the dash0 mono repo and publish them to the public ghcr.

Initially, I had planned to build and push the IE collector image in the dash0 repo, but I think that will increase the chance of version drift between the IE and non-IE collector (wrt otel components). The more idiomatic way would be to have a separate builder config in the operator repo that references the private modules from the dash0 repo (just like we do for all other components we use from the contrib repo). Keeping the IE components updated would then also work via dependabot (once the IE components start to use proper semver).

The downside is, that this will require a token to read from that private dash0 repo, unless we publish the modules somewhere else.

internal/controller/sampling_rule_controller.go
```diff
@@ -0,0 +1,488 @@
...
+		if transformErr != nil {
+			logger.Error(transformErr, "error transforming sampling rule to API payload")
+			return NewResourceToRequestsResultSingleItemError(apiConfig, itemName, transformErr.Error())
+		}
+		serialized, _ := json.Marshal(apiPayload)
+		method = http.MethodPut
+		req, err = http.NewRequest(method, samplingRuleUrl, bytes.NewBuffer(serialized))
+	case deleteAction:
+		method = http.MethodDelete
+		req, err = http.NewRequest(method, samplingRuleUrl, nil)
```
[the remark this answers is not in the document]
p-d0: @AikidoSec ignore: This is not external user input. Both CRs require Kubernetes RBAC (cluster-admin or namespace-admin privileges) to create/modify the CR. Someone with that level of access already has far more powerful attack vectors than SSRF.

internal/controller/sampling_rule_controller.go
```diff
@@ -0,0 +1,488 @@
...
+	switch action {
+	case upsertAction:
+		apiPayload, transformErr := transformToSamplingApiPayload(preconditionChecksResult.resource, itemName, apiConfig.Dataset)
+		if transformErr != nil {
+			logger.Error(transformErr, "error transforming sampling rule to API payload")
+			return NewResourceToRequestsResultSingleItemError(apiConfig, itemName, transformErr.Error())
+		}
+		serialized, _ := json.Marshal(apiPayload)
+		method = http.MethodPut
+		req, err = http.NewRequest(method, samplingRuleUrl, bytes.NewBuffer(serialized))
```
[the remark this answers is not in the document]
p-d0: @AikidoSec ignore: This is not external user input. Both CRs require Kubernetes RBAC (cluster-admin or namespace-admin privileges) to create/modify the CR. Someone with that level of access already has far more powerful attack vectors than SSRF.

helm-chart/dash0-operator/templates/operator/deployment-and-webhooks.yaml
```diff
@@ -163,6 +163,7 @@ spec:
         - --operator-configuration-auto-monitor-namespaces-label-selector={{ .Values.operator.autoMonitorNamespaces.labelSelector }}
 {{- end }}{{/*closes "if .Values.operator.dash0Export.enabled" */}}
         - --operator-configuration-telemetry-collection-enabled={{ .Values.operator.telemetryCollectionEnabled }}
+        - --feature-intelligent-edge-enabled={{ include "dash0-operator.development.intelligentEdge.enabled" . }}
```
basti1302: *issue(blocking):* We will run into issues with the GKE allowlists here. And since the option is added unconditionally, all GKE AutoPilot users would be affected, once this is merged and released.

I think we should only add this command line parameter if `dash0-operator.development.intelligentEdge.enabled` is `true`, and otherwise just omit it.

basti1302: Assuming we implement the change above (adding the parameter conditionally), we then still need to figure out how to allow using IE on GKE Autopilot, but that is a less pressing concern right now.

We could:
1. rename this to `--dash0-feature-intelligent-edge-enabled`. Everything prefixed with `--dash0-` is allowed once https://github.com/dash0hq/dash0-operator/pull/941 is merged (realistically, this might take another 2-3 weeks)
2. extend the currently open merge request for the Dash0 GKE Autopilot allowlists to also allow everything prefixed with `--feature`.

I would rather lean towards (1), but I'm fine with (2) as well if you have strong opinions about the naming. 

Either way, after doing either (1) or (2), and once PR #941 has been released, we can then also remove the conditional again, and always add this command line parameter unconditionally, if that is preferred.

p-d0: Good point. Option 1 and omitting when empty sounds fine :+1:

internal/collectors/collector_manager.go
```diff
@@ -225,6 +238,30 @@ func (m *CollectorManager) findOperatorConfigurationResource(
...
+		ctx,
+		m.Client,
+		"",
+		&dash0v1alpha1.Dash0IntelligentEdge{},
+		logger,
+	)
+	if err != nil {
+		if meta.IsNoMatchError(err) {
+			// The Dash0IntelligentEdge CRD is not installed (IE feature is disabled). This is expected.
+			return nil, nil
```
basti1302: _question:_ Why would that be expected here? If I read this correctly, we only call this if the feature is enabled, so wouldn't the CRD not being installed be an unexpected error here?

internal/collectors/otelcolresources/desired_state.go
```diff
@@ -928,6 +941,15 @@ func assembleCollectorDaemonSetVolumes(
...
 	}
 
+	if config.IntelligentEdge.Enabled && config.IntelligentEdge.SamplingEnabled {
+		volumes = append(volumes, corev1.Volume{
+			Name: "trace-reservoir",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		})
+	}
```
basti1302, line 951: _note:_ We will need to add the new volume and volume mount also to the allowlist at some point, plus allowlists for barker. (No action required now.)

internal/collectors/otelcolresources/desired_state.go
```diff
@@ -928,6 +941,15 @@ func assembleCollectorDaemonSetVolumes(
 		})
 	}
 
+	if config.IntelligentEdge.Enabled && config.IntelligentEdge.SamplingEnabled {
+		volumes = append(volumes, corev1.Volume{
+			Name: "trace-reservoir",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
```
basti1302, line 948: _question:_ Should we provide a sizeLimit? (Not sure if we can come up with a reasonable upper limit though.)

internal/intelligentedge/intelligentedge_manager.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/dash0hq/dash0-operator/internal/intelligentedge/ieresources"
+	"github.com/dash0hq/dash0-operator/internal/resources"
+	"github.com/dash0hq/dash0-operator/internal/util/logd"
+)
+
+type IntelligentEdgeReconcileTrigger string
+
+const (
+	TriggeredByWatchEvent             IntelligentEdgeReconcileTrigger = "watch"
+	TriggeredByDash0ResourceReconcile IntelligentEdgeReconcileTrigger = "resource"
```
basti1302: _nit:_ `TriggeredByDash0ResourceReconcile` is currently unused, I think it could be removed.

Plus, if there is only one possible trigger value, we might consider removing the trigger parameter on the ReconcileIntelligentEdge function, not sure. (Looks like it is only logged but does not influence the actual behavior of `ReconcileIntelligentEdge`.)

test-resources/intelligentedge/README.md
```diff
@@ -0,0 +1,11 @@
+# Temporary helper script to build and push IE images
+
+0. clone the dash0 mono repo
+0. run `DASH0_REPO_ROOT=/path/to/cloned/repo ./build-ie-and-barker.sh --push`
```
basti1302, line 4: We could call out here that it defaults to building amd64 images by default (at least for barker), and that you can override it with ARCH=arm64. (Took me a bit to figure out the ImagePull errors for barker :-) )

test-resources/intelligentedge/build-ie-and-barker.sh
```diff
@@ -0,0 +1,90 @@
...
+cd "$BARKER_DIR"
+CGO_ENABLED=0 GOOS=linux GOARCH="${ARCH}" go build \
+  -ldflags="$LDFLAGS_BARKER" \
+  -o "bin/dash0-barker_linux_${ARCH}"
+
+echo "==> Building operator-intelligent-edge-collector Docker image"
+cd "$COLLECTOR_DIR"
+cp -f "bin/operator-intelligent-edge-collector_linux_${ARCH}" "operator-intelligent-edge-collector_${ARCH}"
+cp -f "$SCRIPT_DIR/entrypoint-operator-intelligent-edge.sh" entrypoint-operator-intelligent-edge.sh
+docker build --build-arg "TARGETARCH=${ARCH}" -t intelligent-edge-collector:latest \
```
basti1302: _note:_ This `docker build` misses `--platform "linux/${ARCH}"` I think. When being run on arm64, and no ARCH is specified, it will compile for amd64, but build a container image tagged as arm64.

test-resources/intelligentedge/build-ie-and-barker.sh
```diff
@@ -0,0 +1,90 @@
...
+  -f "$SCRIPT_DIR/Dockerfile.operator-intelligent-edge" .
+rm -f "operator-intelligent-edge-collector_${ARCH}" entrypoint-operator-intelligent-edge.sh
+
+echo "==> Building barker Docker image"
+cd "$BARKER_DIR"
+cp -f "bin/dash0-barker_linux_${ARCH}" "dash0-barker_${ARCH}"
+docker build --platform "linux/${ARCH}" --build-arg "TARGETARCH=${ARCH}" -t dash0-barker:latest .
```
basti1302: The image is tagged dash0-barker here and also in line 71. 
When pushing images in line 81, and also the default for the image name when deploying intelligent edge seems to be just `barker`, without the `dash0-` prefix.

test-resources/intelligentedge/build-ie-and-barker.sh
```diff
@@ -0,0 +1,90 @@
...
+
+echo "==> Building operator-intelligent-edge-collector binary"
+cd "$COLLECTOR_DIR"
+go tool go.opentelemetry.io/collector/cmd/builder \
+  --skip-compilation \
+  --config "$RESOLVED_CONFIG" \
+  --output-path dist_operator_intelligent_edge_collector
+rm -f "$RESOLVED_CONFIG"
+cd dist_operator_intelligent_edge_collector
+GOWORK=off CGO_ENABLED=0 GOEXPERIMENT=newinliner go build -trimpath \
```
basti1302:
```suggestion
GOWORK=off CGO_ENABLED=0 GOOS=linux GOARCH="${ARCH}" GOEXPERIMENT=newinliner go build -trimpath \
```

Without `GOOS=linux GOARCH="${ARCH}"` this will build a Dwarf/MacOS binary on MacOS and then include it in the linux container image.

internal/intelligentedge/ieresources/desired_state.go
```diff
@@ -0,0 +1,312 @@
...
+		Replicas: &replicas,
+		Selector: &metav1.LabelSelector{
+			MatchLabels: barkerMatchLabels,
+		},
+		Template: corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: barkerLabels(),
+			},
+			Spec: corev1.PodSpec{
+				AutomountServiceAccountToken: ptr.To(false),
```
basti1302: _question:_ Should we also specifiy a SecurityContext?

test-resources/intelligentedge/extra-values.yaml
```diff
@@ -0,0 +1,4 @@
+operator:
+  development:
+    intelligentEdge:
+      enabled: true
```
basti1302: _q:_ Is this a left-over from earlier experiments? It seems it is not referenced anywhere.

basti1302 (approved):
Great work! LGTM, assuming https://github.com/dash0hq/dash0-operator/pull/958/changes#r3108960660 (already discussed) has been fixed.

test-resources/intelligentedge/extra-values.yaml
p-d0: it was used for manual tests but I removed it

test-resources/intelligentedge/README.md
p-d0: changed it to be auto-detected instead of defaulting to amd64

internal/collectors/otelcolresources/desired_state.go
p-d0: I have added a ticket to review the trace reservoir config. There are also some advanced settings that might be worth customizing or exposing.

PR #961, opened 2026-04-19
feat: IaC support for notification channels

mmanciop:


internal/controller/notification_channel_controller.go
```diff
@@ -0,0 +1,599 @@
...
+		// - move spec.display.name → metadata.name (API expects display name there)
+		// - assemble the type-specific config (e.g. spec.slackConfig) into spec.config
+		resource := preconditionChecksResult.resource
+		prepareNotificationChannelApiPayload(resource)
+		serializedResource, _ := json.Marshal(resource)
+		requestPayload := bytes.NewBuffer(serializedResource)
+		method = http.MethodPut
+		req, err = http.NewRequest(
+			method,
+			notificationChannelUrl,
```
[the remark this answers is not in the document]
mmanciop: This is the same pattern used in all other controllers

api/operator/v1beta1/dash0notificationchannels_types.go
```diff
@@ -0,0 +1,332 @@
+// SPDX-FileCopyrightText: Copyright 2026 Dash0 Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1beta1
```
p-d0, line 4: question: all other sync resources are in `v1alpha1`, do we really want the notification channel in `v1beta1`?

mmanciop: What's the sense to go back? When we cut v1beta1, it came with expectations that v1alpha1 would not be regularly updated.

If anything, my counter question os what are we waiting for to promote the other CRDs ;-)

p-d0: Mostly for consistency :) Basically everything is `v1alpha1` atm with the exception of the monitoring resource and afaik that was only because of a breaking change that was introduced...

I'm not against promoting everything.

p-d0 (approved):

PR #973, opened 2026-04-23
fix: race condition in sync controllers

p-d0:
There was a potential race condition since the `defer` released the lock as soon as the go routine was entered and there was a brief period that allowed a second call to run past the `if r.initialSyncHasHappened.Load()` check before the sync had happened (so the flag wasn't true yet).

This triggered for sampling rules more frequently than the other sync resources and caused the tests to be flaky.

## Summary by Aikido
|  Security Issues: 0 |  Quality Issues: 0 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**🐛 Bugfixes**
* Added atomic in-progress flag to prevent initial sync race

<sup>[More info](https://app.aikido.dev/featurebranch/scan/107468256?groupId=48107)</sup>


basti1302 (approved):
👍

PR #979, opened 2026-04-24
feat: collect barker metrics and traces

p-d0:


internal/collectors/otelcolresources/daemonset.config.yaml.template
```diff
@@ -834,6 +834,16 @@ processors:
...
+  {{- if and .SelfMonitoringEnabled .IntelligentEdge.Enabled .IntelligentEdge.BarkerEnabled }}
+  # Barker's stdout logs are already collected by filelog/selfmonitoring, but the shared OpenTelemetry init helper also
+  # installs an OTLP log exporter whenever OTEL_EXPORTER_OTLP_ENDPOINT is set (no per-signal toggle), so barker also
+  # pushes its logs via OTLP. Drop those SDK-originated records to avoid duplicates. telemetry.sdk.language is set by
+  # the OTel SDK and absent on filelog-scraped records.
+  filter/logs/drop_barker_otlp_selfmonitoring:
+    logs:
+      log_record:
+        - 'resource.attributes["service.name"] == "barker" and resource.attributes["service.namespace"] == "dash0-operator" and resource.attributes["telemetry.sdk.language"] != nil'
+  {{- end }}
```
basti1302, line 846: _thought:_ I guess we could also make `images/pkg/common/otel.go#InitOTelSdkFromEnvVars` smarter and react to OTEL_LOGS_EXPORTER=none accordingly instead of having barker export OTLP logs and then dropping them again? Not saying that we need to do that, or that we need to do that right now. Both approaches are fine IMO either way.

basti1302 (approved):
👍

internal/collectors/otelcolresources/daemonset.config.yaml.template
p-d0: I think this is something to be improved in `backend-utils` but I don't think it will be considered high priority. So for now this seemed the best approach, unless you disagree (maybe I misread your message but I do not see how this could be addressed in this PR).

basti1302: Ah, right, my bad. 🤦  I was somehow thinking we are using our own shared OTel package here (the one from the operator repo), but that's of course not the case. Thanks for the explanation. 👍

PR #981, opened 2026-04-25
chore: check API endpoint against allowed hosts

basti1302:


basti1302:
This should address the "HTTP request might enable SSRF attack" finding.

p-d0 (commented):
This input is already protected by k8s RBAC and I disagree with making changes just because a tool (Aikido) creates spam.

Please go ahead and merge it without my approval.

basti1302:
> This input is already protected by k8s RBAC and I disagree with making changes just because a tool (Aikido) creates spam.

Fair point, then we'll just continue to dismiss these findings when introducing new IaC capabilities.

PR #991, opened 2026-04-28
test(e2e): add e2e tests for initial IE components

p-d0:
## Overview

- adds `control-plane` and `decision-maker` mock services
- adds e2e tests that verify things are wired together correctly
  - check barker is deployed and IE features are enabled when the feature flag is enabled
  - check barker talks to decision maker
  - check dash0settings talks to the control plane
  - check RED metric `dash0.spans.red` is sent
  - check cleanup logic (when IE is disabled again)

## Additional changes

- adds an option to disable Slack notifications for the e2e workflow so that repeated runs don't spam the channel


test/e2e/control-plane-mock/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# SPDX-FileCopyrightText: Copyright 2026 Dash0 Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+FROM --platform=${BUILDPLATFORM} golang:1.26.3-alpine3.23 AS builder
```
basti1302, line 4: q: should we add the parent directory of this file to the docker group in test/e2e/control-plane-mock/Dockerfile?

basti1302: probably also to the gomod group that updates the go.mod files of the other e2e apps.

basti1302 (approved):
👍

PR #994, opened 2026-04-29
chore(collectors): rename filelog receiver to file_log

basti1302:
Starting with OpenTelemetry collector release 0.149.0, the filelog receiver has been renamed to file_log, with filelog becoming a deprecated alias.

This commit introduces the new name in the collector configuration templates.

- the receiver is referred to by "file_log" now
- named pipelines referring to that name use "file-log", since they use hyphens as separators (e.g. logs/file-log-to-forwarder)
- the filelog offset container images and containers or anything related to the offset mechanism has not been renamed

Unfortunately this introduces a mix of file_log, file-log and filelog for now, but renaming the offset-related container images and containers is out of scope for this commit.

### DIFF
--- a/internal/collectors/otelcolresources/daemonset.config.yaml.template
+++ b/internal/collectors/otelcolresources/daemonset.config.yaml.template
@@ -507,7 +507,7 @@ receivers:
   {{- end }}
 
   {{- if .NamespacesWithLogCollection }}
-  filelog:
+  file_log:
     include:
     {{- range $i, $namespace := .NamespacesWithLogCollection }}
       - /var/log/pods/{{ $namespace }}_*/*/*.log
@@ -527,7 +527,7 @@ receivers:
   target-allocator is running even though it shouldn't.
   */}}
   {{- if and .SelfMonitoringEnabled (or .PrometheusCrdSupportEnabled (and .IntelligentEdge.Enabled .IntelligentEdge.BarkerEnabled)) }}
-  filelog/selfmonitoring:
+  file_log/selfmonitoring:
     include:
       {{- if .PrometheusCrdSupportEnabled }}
       - /var/log/pods/{{ .OperatorNamespace }}_{{ .OperatorResourcesNamePrefix }}-opentelemetry-target-allocator*/*/*.log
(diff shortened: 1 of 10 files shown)


internal/collectors/otelcolresources/daemonset.config.yaml.template
```diff
@@ -1329,19 +1329,19 @@ service:
...
+        - file_log
       processors:
         - memory_limiter
       exporters:
         - forward/logs-processors
     {{- end }}
 
     {{- if and .SelfMonitoringEnabled (or .PrometheusCrdSupportEnabled (and .IntelligentEdge.Enabled .IntelligentEdge.BarkerEnabled)) }}
-    logs/selfmonitoring-filelog-to-forwarder:
+    logs/selfmonitoring-file-log-to-forwarder:
```
p-d0: minor nit: I think keeping `filelog` would have been fine here, we also consistently use `self-monitoring` in the [docs](https://github.com/dash0hq/dash0-operator/tree/main/helm-chart/dash0-operator#disable-self-monitoring) and here we use it without a hyphen.

It's also still written as one word in `transform/logs/filelog_service_attributes`.

p-d0 (approved):

internal/collectors/otelcolresources/daemonset.config.yaml.template
basti1302: I agree that using filelog for that is probably better; at least we do not have a mix of three different things then. 👍  Will change it.

basti1302:
> ### ❌ E2E tests failed
> * **Branch**: `collector-config-updates-file_log-receiver`
> * **Focus**: `log collection`
> * **Run**: [#25125577776](https://github.com/dash0hq/dash0-operator/actions/runs/25125577776)

```
Dash0 Operator with an existing operator deployment and operation configuration resource with a deployed Dash0 monitoring resource log collection [It] toggles OTEL_LOGS_EXPORTER=none with the log collection setting
/home/runner/work/dash0-operator/dash0-operator/test/e2e/e2e_test.go:368

  [FAILED] Timed out after 60.001s.
  The function passed to Eventually failed at /home/runner/work/dash0-operator/dash0-operator/test/e2e/pod_env.go:67 with:
  env var OTEL_LOGS_EXPORTER on deployment "dash0-operator-jvm-spring-boot-test-deployment", container "app"
  Expected
      <string>: 
  to equal
      <string>: none
  In [It] at: /home/runner/work/dash0-operator/dash0-operator/test/e2e/e2e_test.go:435 @ 04/29/26 18:17:02.038
```

I think this is unrelated to the changes in this PR, my guess is that this test is flaky when running the test suite with `GINKGO_FOCUS=log collection`. Not sure though.

PR #1005, opened 2026-05-04
test(iac): fix flaky tests

basti1302:
The test "deletes check rules with two API configs (multi-export)" bled over into "updates check rules", after introducing IaC support for recording rules in commit 856ae9e7f37d182be9c98a633a1f9624e553f0cb.

The delete test was missing gock mocks for new HTTP delete calls introduced for recording rules, hence these failed with "gock: cannot match any request", without failing the test. The next test case would already start, while the goroutine trying to delete the recording rule and writing the monitoring resource status were still running.

Fixed by:
- add missing gock expectations
- wait for the status to be written to the monitoring resource at the end of the test
- verify there are no unmatched gock requests after each test

### DIFF
--- a/internal/controller/notification_channel_controller_test.go
+++ b/internal/controller/notification_channel_controller_test.go
@@ -84,6 +84,8 @@ var _ = Describe(
 
 		AfterEach(
 			func() {
+				VerifyNoUnmatchedGockRequests()
+
 				DeleteMonitoringResource(ctx, k8sClient)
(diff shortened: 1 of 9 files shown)


basti1302:
With these fixes, running either the Prometheus rule or the Perses dashboard unit tests in a loop works fine for 100 iterations. On main, I could not get past ~10 iterations for the Prometheus rule tests.

p-d0 (approved):

PR #1013, opened 2026-05-07
ci: allow running the GKE AP allowlist check for a PR/branch

basti1302:


basti1302:
Merging this improved CI action without review since I want to have it available in #941.

PR #1014, opened 2026-05-07
feat(workloads): don't instrument workloads w. low storage limits

basti1302:
Adding the dash0-instrumentation volume to workloads with tight ephemeral storage limits will cause pod eviction.


p-d0 (approved):

basti1302:
I think I'd like to merge this one after #1009 actually, and then only reject instrumenting a container with low storage limits if the instrumentation delivery is != image volume.

basti1302:
> I think I'd like to merge this one after #1009 actually, and then only reject instrumenting a container with low storage limits if the instrumentation delivery is != image volume.

Added a blocked label to make it more obvious that this one is currently on hold.

PR #1019, opened 2026-05-08
chore: update IE image helpers

p-d0:
- builds the IE images using the files from the source repo
- adds make targets for easier local building


p-d0:
merging this since it's just non-production utility code and is required to continue with #991

PR #1025, opened 2026-05-12
ci: skip Slack notification for e2e tests when triggered via PR comment

basti1302:


p-d0:
I think this would currently lead to an incorrect/surprising behavior when running with the settings below. No Slack notification would be sent despite the checkbox saying otherwise.

e2e_settings

Wouldn't it be better to use the input param to disable the Slack notification here: https://github.com/dash0hq/dash0-operator/blob/main/.github/workflows/pr-comment-e2e-trigger.yaml#L91

basti1302:
> I think this would currently lead to an incorrect/surprising behavior when running with the settings below. No Slack notification would be sent despite the checkbox saying otherwise.

I would not expect anyone to manually set the pr_number property when starting the tests manually. The pr_number is also used as a signal that this was kicked off via a PR comment in the "Post PR comment" actions. But sure, we can also use the new notify_slack property.

p-d0:
> I would not expect anyone to manually set the pr_number property when starting the tests manually.



> > I think this would currently lead to an incorrect/surprising behavior when running with the settings below. No Slack notification would be sent despite the checkbox saying otherwise.
> 
> I would not expect anyone to manually set the pr_number property when starting the tests manually. The pr_number is also used as a signal that this was kicked off via a PR comment in the "Post PR comment" actions. But sure, we can also use the new notify_slack property.

in that case, we should remove the input from the `workflow_dispatch`

then this approach is also fine if you prefer it

p-d0 (approved):

basti1302:
> in that case, we should remove the input from the `workflow_dispatch`

pr-comment-e2e-trigger [uses](https://github.com/dash0hq/dash0-operator/blob/af84ef9fa403b32211da9d241bf71c3373bd67df/.github/workflows/pr-comment-e2e-trigger.yaml#L95-L101) workflow dispatch to trigger the e2e tests from a PR comment, and passes the pr_number parameter. I would not expect someone to set it when triggering e2e tests _manually_.

Edit: And one reason e2e.yaml is triggered via workflow dispatch instead of workflow call, is that it allows using the workflow definition of e2e.yaml from the PR branch. A workflow call would use the one from `main`. Using the definition from the branch is rather handy for testing changes in e2e.yaml, using the definition from `main` might be surprising and unintuitive.

PR #1038, opened 2026-05-16
feat(operator): identify outbound API traffic with a User-Agent header

mmanciop:
## Summary

Today the operator sets no `User-Agent` on requests to the Dash0 API — calls go out with Go's default `Go-http-client/1.1`, indistinguishable from any other Go program. This PR adds an identifying `User-Agent` of the form `Dash0 Operator/<version>` so backend logs and analytics can attribute traffic to the operator and the operator version.

## Approach

- New `internal/util/http_useragent.go`:
  - `RenderUserAgent(version)` returns `Dash0 Operator/<version>`, falling back to `Dash0 Operator/unknown` when the version is empty (e.g. the operator image carries neither a tag nor a digest).
  - `WithUserAgent(client, version)` wraps an `*http.Client` with a small `http.RoundTripper` that injects the header on outgoing requests, unless the caller has already set a `User-Agent`. The wrapper clones the request before mutation so the caller's header map is left untouched. The original `Timeout`, `CheckRedirect`, and `Jar` are preserved.
- `internal/startup/operator_manager_startup.go` wraps the shared `httpClient` (which all API-sync reconcilers — synthetic checks, views, notification channels, spam filters, sampling rules, Perses dashboards, Prometheus rules — receive) once at startup, right after `images` is built, so the version from `images.GetOperatorVersion()` is in scope.

The wrapper sits outside the existing retry/rate-limit transport from `dash0-api-client-go`, so the header is set once and naturally rides along on retried requests as well.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./internal/...` clean
- [x] `golangci-lint run ./internal/util/... ./internal/startup/...` — 0 issues
- [x] New Ginkgo specs in `internal/util/http_useragent_test.go` (7 specs) all pass — cover the render fallback, header injection against an `httptest.Server`, preservation of caller-set User-Agent, non-mutation of the caller's request, and timeout passthrough.
- [ ] CI: `make test` (controller/startup suites need envtest binaries that aren't available in the local container; verifying on CI)

---
_Generated by [Claude Code](https://claude.ai/code/session_01JukJMkg23R8pGugAaJiYvr)_

### DIFF
--- a/internal/startup/operator_manager_startup.go
+++ b/internal/startup/operator_manager_startup.go
@@ -199,11 +199,6 @@ var (
 	extraConfig                     util.ExtraConfig
 	extraConfigMapWatcher           = util.NewExtraConfigWatcher()
 
-	httpClient = dash0apiclient.NewTransport(
-		dash0apiclient.WithTransportMaxRetries(2),
-		dash0apiclient.WithTransportRetryWaitMin(1*time.Second),
-		dash0apiclient.WithTransportRetryWaitMax(3*time.Second),
-	).HTTPClient()
 	thirdPartyResourceSynchronizationQueue *workqueue.Typed[controller.ThirdPartyResourceSyncJob]
 )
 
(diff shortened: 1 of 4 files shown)


internal/startup/operator_manager_startup.go
```diff
@@ -197,11 +197,6 @@ var (
 	extraConfig                     util.ExtraConfig
 	extraConfigMapWatcher           = util.NewExtraConfigWatcher()
 
-	httpClient = dash0apiclient.NewTransport(
-		dash0apiclient.WithTransportMaxRetries(2),
-		dash0apiclient.WithTransportRetryWaitMin(1*time.Second),
-		dash0apiclient.WithTransportRetryWaitMax(3*time.Second),
-	).HTTPClient()
```
basti1302, line 206: removed as global var to avoid declaring it without user agent first, and then redefining it with the user agent wrapper later

internal/util/http_useragent.go
```diff
@@ -1,28 +1,31 @@
...
 import (
 	"net/http"
+	"strings"
 )
 
 const (
-	UserAgentHeaderName = "User-Agent"
+	userAgentHeaderName = "User-Agent"
 
-	userAgentProduct        = "Dash0 Operator"
```
basti1302: I think the product name should not have a whitespace, see https://datatracker.ietf.org/doc/html/rfc7231#section-5.5.3 and https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6.

basti1302 (approved):
LGTM

basti1302:
@mmanciop FYI this is currently blocked by the CLA issue due to the commit being authored by Claude. Once this is sorted out, we can merge it.

PR #1062, opened 2026-05-22
feat(auto-monitoring): (re)create auto-resource when it is deleted

basti1302:
This solves two similar but distinct use cases:
- A manually managed Dash0Monitoring resource in a namespace eligible for automatic namespace monitoring blocks creating the auto resource. This is by design. When the manually managed resource is deleted, trigger creating the auto resource immediately.
- An auto-resource is deleted by a user. We now restore it automatically.


basti1302:
~~stacked PR, #1061 needs to be merged first~~ done

p-d0 (approved):

PR #1077, opened 2026-05-28
feat: use pick_first lb for dash0 export and make lb configurable for custom grpc exporters

p-d0:
- uses the `pick_first` balancer for the dash0 exporter since client-side load balancing is not needed and the `round_robin` balancer that the otel collector uses by default since [0.145.0](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.145.0) causes excessive warnings when the dns returns both an A and AAAA record
- for custom grpc exporters the balancer is configurable and by default uses `round_robin` to be consistent with the otel collector


basti1302 (approved):
👍

PR #1117, opened 2026-06-12
chore(collectors): add dash0.metric.datapoint.interval scope attribute

basti1302:


basti1302:
This is blocked until the Dash0 backend correctly processes this new scope attribute. Without that prerequisite, adding this in the operator adds no value.

Issue #1107, opened 2026-06-10
otel-collector daemonset crashes on AWS M8i instances - "failed to shutdown the retiring config"


dwilliams782:
I was on the fence whether to write this up here or in the opentelemetry repo, so we can move it / cross-post.

**Describe the bug**
When a Dash0Monitoring CRD is added / removed, the otel-collector daemonset pods get restarted via config reload. We've had sporadic invalid memory address errors during this reload for some time (which is tracked [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47672)) but is separate to this new issue.

We recently expanded our EKS machine types via Karpenter to allow M8i family machines. Following this, we got spammed with container restarts on the otel-collector daemonset with context deadline exceeded logs:

```
2026-06-10T10:00:06.013Z	info	service@v0.151.0/service.go:292	Shutdown complete.	{"resource": {"k8s.cluster.name": "x", "k8s.cluster.uid": "x", "k8s.container.name": "opentelemetry-collector", "k8s.daemonset.name": "dash0-operator-opentelemetry-collector-agent-daemonset", "k8s.daemonset.uid": "86564bc5-8817-4449-aa52-64d9cf99cbd1", "k8s.namespace.name": "dash0-system", "k8s.node.name": "x", "k8s.pod.name": "dash0-operator-opentelemetry-collector-agent-daemonset-b8m2v", "k8s.pod.uid": "96683afb-df95-4e51-9da6-492b8ec3222b", "service.instance.id": "4fd57850-7aed-45aa-9fa0-d853eb744ebd", "service.name": "dash0-operator-collector", "service.version": "0.151.0"}}
Error: failed to shutdown the retiring config: failed to shutdown logger: context deadline exceeded: rpc error: code = DeadlineExceeded desc = context deadline exceeded
``` 

I reverted the Karpenter change to remove M8i machines, and these ContextDeadlineExceeded errors stopped. 

I then added a new debug Karpenter nodepool of M8i, running a single node - the rest of the cluster is on 7th generation or lower, and during config reloads, only the M8i pod restarts:

```
k get pods | grep agent
dash0-operator-opentelemetry-collector-agent-daemonset-8tsnd      2/2     Running     0               63m
dash0-operator-opentelemetry-collector-agent-daemonset-b8m2v      2/2     Running     5 (5m34s ago)   100m
dash0-operator-opentelemetry-collector-agent-daemonset-h7lpc      2/2     Running     0               88m
[...]

PR #1120, opened 2026-06-16
feat: allow setting pod sysctls via helm

p-d0:
Allows configuring sysctls for collector pods via helm to lower the TCP keepalive if needed (e.g. on AWS Nitro V6 instances). The issue is described in greater detail in #1107.

Follows the suggestion under `TCP keepalives` at https://aws.amazon.com/blogs/networking-and-content-delivery/best-practices-for-tcp-connection-management-on-ec2/

The recommended sysctls are classified as safe since k8s 1.29: https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#safe-and-unsafe-sysctls

Note: `exporters` in the collector config support keepalive directly, but the self-telemetry export does not. See https://opentelemetry.io/docs/specs/otel/configuration/types/ (and search for `OtlpGrpcExporter`) vs https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlpexporter/README.md


helm-chart/dash0-operator/values.yaml
```diff
@@ -779,6 +795,19 @@ operator:
         memory: 12Mi
         # ephemeral-storage: (no ephemeral-storage request by default)
 
+    # An array of pod-level sysctls applied to the collector Deployment pod (cluster-metrics-collector), see
+    # spec.securityContext.sysctls. Disabled by default. See daemonSetSysctls above for the use case
+    # (forcing TCP keepalive below the connection-tracking idle timeout, e.g. on AWS Nitro v6 nodes) and the Kubernetes
+    # >= 1.29 / Pod Security Admission caveats.
+    # Example:
+    # deploymentSysctls:
```
basti1302, line 803: _q:_ Is there a use case for having different sysctls for the daemonset and the deployment? Or would we expect that they are usually identical? I wonder if we should just offer one setting that is applied uniformly to both.

helm-chart/dash0-operator/README.md
```diff
@@ -1502,6 +1502,55 @@ operator:
...
+```
+
+`net.ipv4.tcp_keepalive_time` is the idle time before the first keepalive probe is sent and must be lower than the
+environment's conntrack idle timeout (for example, below 350 seconds on AWS Nitro v6); `tcp_keepalive_intvl` and
+`tcp_keepalive_probes` only govern dead-peer detection.
+
+This setting is opt-in for two reasons:
+
+* The `net.ipv4.tcp_keepalive_*` sysctls are only considered _safe_ sysctls on Kubernetes 1.29 and later. On older
+  clusters, or on clusters where the kubelet's `forbiddenSysctls` / a strict Pod Security Admission policy disallows
```
basti1302, line 1547: Note: We usually know the K8s version because `operator_manager_startup` calls `cluster.DetectKubernetesVersion`. Does that change the picture?

Should we consider setting the tcp_keepalive_* sysctls automatically? We could probably also find out whether this is on AWS and what instance type it is.

I'm not necessarily saying we should, only that it's worth considering. WDYT?

basti1302 (approved):
👍 👍

helm-chart/dash0-operator/README.md
p-d0: I'd prefer to keep it simple for the following reasons:

- some people might prefer to fix the issue by configuring the TCP keepalive via a `NodeClass` instead

- >  We could probably also find out whether this is on AWS and what instance type it is.

   Most people will probably use a mix of different instances. And the instances could change after the operator has been installed.

helm-chart/dash0-operator/values.yaml
p-d0: Mostly for consistency. I'd expect them to be the same most of the time unless someone uses a different node affinity for the deployment and it's therefore not needed there.

basti1302: Also really no strong opinion here. If there is no use case for having both set differently, I would tend to not force users to set this in two places. If there is a potential use case, having independent settings is fine. Sounds like the example with different node affinities is a use case?

basti1302: About consistency: We already have a couple of settings that can only be set once, but affect both workloads (debugVerbosityDetailed, sendBatchSize, enablePprofExtension), so having only one setting would not be entirely inconsistent.

helm-chart/dash0-operator/README.md
basti1302: Good points 👍

helm-chart/dash0-operator/values.yaml
p-d0: yes, but that is a different category of settings (imho) - in the same category (pod-level) we have 
- `deploymentTolerations`/`daemonSetTolerations`
- `deploymentProbes`/`daemonSetProbes`
- `deploymentNodeAffinity`/`daemonSetNodeAffinity`

so the `deploymentSysctls`/`daemonSetSysctls` seemed the most natural structure

but I agree there's a less clear/likely use case to require different settings for deployment/daemonset compared to the others

basti1302: FWIW, I'm fine with either structure. 👍

PR #1124, opened 2026-06-17
chore(gke): bump gke allowlist labels to 1.0.4

p-d0:
this should only be merged after the referenced allowlist 1.0.4 is available


basti1302:
Note: We still need to update https://github.com/dash0hq/dash0-operator/blob/a7d8e9cb21dc421d46a901ceda072ea22043ee92/helm-chart/dash0-operator/templates/operator/pre-install-hook-allowlist-synchronizer-ready-check.yaml#L105 to 1.0.4, but that can only be done after the new version has been published by Google, otherwise it will block the testing workflow.

PR #1134, opened 2026-06-24
chore(agent0-connector): add GKE AutoPilot allowlist label

basti1302:


basti1302:
~~#1123 needs to be merged before this~~ ✅

basti1302:
I'm going to fold this one into #1124, so we have all the allowlist changes in one PR.

PR #1143, opened 2026-06-26
fix: remove unused trace propagator packages for Python

mmanciop:
I have not come across anybody using B3 or Jaeger propagation headers in this decade


basti1302 (approved):

basti1302:
FWIW, these came from here https://github.com/open-telemetry/opentelemetry-operator/blob/572831511fa44a66e910d3032c540f45ee3439ce/autoinstrumentation/python/requirements.txt#L9-L10. But +1 on removing them.

PR #1144, opened 2026-06-26
ci: update docs sync after restructuring

basti1302:
* update .github/workflows/sync-docs-to-website.yaml and .github/workflows/sync-docs to work with multiple pages instead of one monolithic README.md file
* rename docs/*.md pages to use lower case file names
* use page or section names in relative links as the link text
* fix a couple of broken links


basti1302:
~~requires #1131~~

PR #1178, opened 2026-07-07
fix(perses): preserve panel-spec unknown fields in PersesDashboard CRD

duart38:
The upstream PersesDashboard CRD models the per-panel `spec` as a closed schema (only `display`/`links`/`plugin`/`queries`) with no `x-kubernetes-preserve-unknown-fields`. Dash0 panel extensions at panel-spec level (e.g. the per-panel `annotations` block) are pruned by the API server at admission, before the operator reads the resource, so they never reach Dash0.

The operator now patches the CRD to set `x-kubernetes-preserve-unknown-fields: true` at the panel spec for every version, so these extensions survive. Behavior:

- Gated by the new Helm value `operator.iac.persesDashboard.autoPreservePanelSpecFields` (default `true`).
- Skipped when a foreign conversion webhook is present (the Perses operator owns the CRD); in that case the durable fix is upstream Perses or managing the dashboard via the Dash0 API/UI.
- Reuses the existing CRD reconcile path alongside the conversion-webhook patch.

Tests: envtest reproduces the pruning against the unpatched CRD and proves the field survives after the patch; plus foreign-webhook backoff and an idempotent unit test for the schema helper.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/helm-chart/dash0-operator/docs/managing-dash0-resources.md
+++ b/helm-chart/dash0-operator/docs/managing-dash0-resources.md
@@ -124,6 +124,22 @@ If you leave `operator.iac.persesDashboard.autoPatchConversionWebhook` enabled a
 `spec.conversion` on every sync, the operator will re-apply the patch on each CRD update event and log a warning, so the
 back-and-forth patching between the Dash0 operator and the GitOps system is observable.
 
+#### Preserving Dash0 Panel Extensions in the PersesDashboard CRD
+
+The upstream `persesdashboards.perses.dev` CRD models the per-panel `spec` as a closed schema (only `display`, `links`,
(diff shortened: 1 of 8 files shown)


duart38:

duart38:
recheck

basti1302 (changes requested):
Discussed offline, conceptually this might not be the direction we want to take here.

Marking with "Request changes" not so much because there are specific changes required for the PR content but because this needs a wider discussion first.

basti1302:
Closing this, since we are taking a different and more sustainable route for this, see https://github.com/perses/perses-operator/issues/435, https://github.com/perses/spec/pull/61.

PR #1196, opened 2026-07-09
feat(sc): restructure pipelines to support namespaced/multiple datasets

p-d0:
Restructures the collector pipelines so that

- the SC components run after the non-SC processors
- the dataset is correctly stamped onto the telemetry before it flows through the SC components (it's required to look up spam filters, s2m rules, sampling rules)
- in case of multiple exporters, tail sampling is only applied to the telemetry on the first exporter branch - this is because of a current limitation in the decision maker backend
- for tail sampling, the dataset is first stamped on and then a single sampling processor is reused - otherwise the memory consumption (for the reservoir) would grow with the number of exporters/datasets

Other minor changes

- ensures the SC components are not inserted on non-Dash0 exporter branches

### DIFF
--- a/internal/collectors/otelcolresources/daemonset.config.yaml.template
+++ b/internal/collectors/otelcolresources/daemonset.config.yaml.template
@@ -21,14 +21,26 @@ use option 2.
 */}}
 
 {{- $hasNamespacedExporters := gt (len .Exporters.Namespaced) 0 -}}
-{{- /* Routing (splitting telemetry per destination) is only used when there are namespaced exporters: the routing
-connector requires a non-empty table, keyed on k8s.namespace.name. $useRouting is an alias for that condition.
-$splitDefault: when a (namespaced) config's DEFAULT export list mixes Dash0 + non-Dash0 exporters with SC enabled,
-the default telemetry is also split — SC'd -> default Dash0 exporters, raw -> default passthrough. Non-namespaced
-configs do not split the default (SC stays inline in common-processors, applied to all default exporters, as before
-this feature); splitting the non-namespaced default would need a fork rather than routing (see note). */}}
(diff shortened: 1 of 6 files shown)


internal/collectors/otelcolresources/daemonset.config.yaml.template
```diff
@@ -81,70 +89,157 @@ connectors:
...
+        condition: attributes["k8s.namespace.name"] == "{{ $ns }}"
+        pipelines:
+        {{- range $b := $sc.DatasetBranches }}
+          - traces/sc/ns/{{ $ns }}/{{ $b.DatasetIndex }}
+        {{- end }}
+        {{- if $sc.Passthrough }}
+          - traces/export/ns/{{ $ns }}/passthrough
+        {{- end }}
+    {{- end }}
+{{- else }}
```
basti1302, line 171: _issue (blocking):_ I think we can reach this if $useRouting is true due to $hasDefaultPassthrough, but there are no namespaced exporters. E.g. with sc enabled, a non-Dash0 default export and no namespaced exporters, this generates an empty table, leading to a collector crash at startup. Same for the other signals and metric routing in the deployment config map.

Admittedly, that would be a rather odd config; a non-Dash0 export with SC enabled makes little sense I guess? But as far as I can tell, we already accommodate for that combination to some degree.

basti1302: Nevermind, already resolved with an additional commit after I wrote the original comment.

basti1302: We could look into whether it would be possible to run the certain parts of the config through `Validate()` of the respective OTel component. (Out of scope for this PR, maybe something for OPE-175.

basti1302 (approved):

PR #1197, opened 2026-07-10
ci(gke ap): run allowlist check on push to main

basti1302:


.github/actions/slack-notification/action.yaml
```diff
@@ -0,0 +1,87 @@
+# SPDX-FileCopyrightText: Copyright 2026 Dash0 Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+name: send Slack notification
```
basti1302, line 4: extracted this into a reusable action

basti1302:
> run-allowlist-check

^ will fail, but I want to double check whether the GH action in general still works correctly, independent of the check's success

basti1302:
> ### ❌ GKE Autopilot WorkloadAllowlist check failed (image build)
> * **Branch**: `run-gke-workload-allowlist-on-main-after-every-push`
> * **Run**: [#29096785859](https://github.com/dash0hq/dash0-operator/actions/runs/29096785859)

That failed in the wrong way due to an unrelated access issue for the new agent0-connector image. Re-running.

basti1302:
> ### ❌ GKE Autopilot WorkloadAllowlist check has failed
> * **Branch**: `run-gke-workload-allowlist-on-main-after-every-push`
> * **Run**: [#29100781521](https://github.com/dash0hq/dash0-operator/actions/runs/29100781521)

This failed in the correct/expected way now (it's a new command line flag for the deployment collector). 👍

basti1302:
Added one more thing (collecting diagnostics on failure), so let's do another test run.

.github/workflows/scripts/gke-ap-workload-allowlist-check.sh
```diff
@@ -27,12 +30,105 @@ use_local_chart="${USE_LOCAL_CHART:-}"
...
+      -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' 2>/dev/null)
+  done
+
+  if tar -czf "$diagnostics_archive" "$diagnostics_dir"; then
+    echo "cluster diagnostics have been collected in \"$diagnostics_archive\""
+  else
+    echo "WARNING: failed to create the diagnostics archive \"$diagnostics_archive\""
+  fi
+  set -e
+}
```
basti1302, line 125: Slightly unrelated to the original intent, but so far the allowlist check only failed, and left us guessing what went wrong. All the relevant logs etc. would have been lost once the GH action reported failure (due to `cleanup`). Collecting evidence before doing helm uninstall makes this much more convenient.

basti1302:
This is ready for review, but we should only merge it after https://github.com/dash0hq/dash0-operator/pull/1124.

p-d0 (approved):

PR #1213, opened 2026-07-14
ci: add dry-run step for docs sync

basti1302:
Verifies that the docs sync will run successfully with the next release.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,11 +6,7 @@ on:
       - 'main'
     tags:
       - '*.*.*'
-    paths-ignore:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -6,11 +6,7 @@ on:
       - 'main'
     tags:
       - '*.*.*'
-    paths-ignore:
-      - '*.md'
   pull_request:
-    paths-ignore:
-      - '*.md'
```
basti1302, line 13: These path-ignores really don't serve any purpose, let's get rid of them.

p-d0 (approved):

Issue #1214, opened 2026-07-14
operator-controller logs the resolved Dash0 auth token in plaintext at info level


lenntt:
**Describe the bug**

The operator-controller logs the *resolved* Dash0 auth token in cleartext at `info` level. On every reconcile of the operator configuration it logs the whole self-monitoring config struct, which includes the resolved token:

- Log call: [`internal/controller/operator_configuration_controller.go#L223`](https://github.com/dash0hq/dash0-operator/blob/b233c41d9340ec3789d96ac2d0f886dc8401c51e/internal/controller/operator_configuration_controller.go#L223)
  ```go
  logger.Info("applying self-monitoring configuration", "configuration", selfMonitoringConfiguration)
  ```
- The struct [`SelfMonitoringConfiguration`](https://github.com/dash0hq/dash0-operator/blob/b233c41d9340ec3789d96ac2d0f886dc8401c51e/internal/selfmonitoringapiaccess/self_monitoring_and_api_access.go#L28-L36) has `Token *string` (L32) with no redacting `LogValue()`/`MarshalLogObject`, so zap emits `"Token":"auth_…"`. `ResolvedSecretHeaderValues` (L35) would leak identically for header-based exports.

Anyone with read access to the operator pod logs — or any log-shipping pipeline — can lift the ingest token. In our setup it was tailed into Loki by a cluster-wide log collector.

**Environment (please complete the following information):**
 - Cloud Provider: Apache CloudStack (self-hosted)
 - Cloud Provider Kubernetes Offering: CloudStack Kubernetes Service (CKS)
 - Kubernetes version: v1.36.0 (not environment-specific — the logging is unconditional)
 - Dash0 operator / chart version: 0.146.0 (also present on `main` @ `b233c41`)

**To Reproduce**

1. Install with a Dash0 export (`operator.dash0Export.enabled=true`, token supplied via `secretRef`).
2. Inspect the operator log:
   ```bash
   kubectl -n dash0-system logs deploy/dash0-operator-controller | grep "applying self-monitoring configuration"
   ```
3. The line contains `"Token":"auth_…"` in cleartext.

**Expected behavior**

The resolved token (and `ResolvedSecretHeaderValues`) never appear in logs — e.g. a redacting `LogValue()` / `MarshalLogObject` on `SelfMonitoringConfiguration`, or omit those fields from the log call.

**Additional context**

[...]

PR #1216, opened 2026-07-15
fix(logging): redact tokens when logging config and resources

p-d0:
- addresses the issue raised in #1214
- also redacts tokens when logging resources, patches, etc... (this only affected clear text tokens, not tokens coming from secrets)
- added a check to e2e tests, so on all paths that we test, we also check that the token wasn't logged


basti1302 (approved):
👍 👍

PR #1220, opened 2026-07-16
ci: use dash0hq/sync-docs-action for docs sync

geertjanw:
**What**

  Migrates the docs sync to the shared dash0hq/sync-docs-action composite action, replacing the in-repo
  Python transformation engine and the hand-rolled PR flow.

  Pinned to [dash0hq/sync-docs-action@v0.4.0](https://github.com/dash0hq/sync-docs-action/releases/tag/v0.4.0), which adds a built-in dry-run mode, a docs-coverage guard, and reviewer/assignee inputs.

  **Why**

  sync-docs-action is the shared implementation every source repo (otel-cicd-action, dash0-sdk-web,
  dash0-operator, …) is meant to consume, so each repo no longer maintains its own copy of the
  transformation engine and PR-opening logic.

  **Changes**

  - Delete the Python engine and its support files under .github/workflows/sync-docs/:
  apply-transformations.py, requirements.txt, test-locally.sh, CLAUDE.md, .gitignore.
  - Rewrite sync-docs-to-website.yaml. Both modes are a single invocation of the composite action —
  dry-run: ${{ inputs.dry-run }} selects between them. On a dry run the action applies the transformations
  - coverage check and stops (no target checkout, no PR, no token); the real sync additionally checks out the target repo, copies the pages, and opens/updates the PR.
  - Fix transformations.yaml. The leading-heading strip used the Python-only \A anchor; JavaScript's RegExp treats \A as a literal A, so the (required) transformation matched nothing under the TypeScript engine. Changed to ^ (equivalent without the multiline flag). Comments referencing the removed Python engine were updated.
  - Add a coverage: guard to transformations.yaml (include: README.md, docs/*.md). Any docs file not wired into files: (or explicitly listed under ignore) now fails the run, including the CI dry-run — restoring the "add a page → CI breaks until it's wired up" safety net.
  - Keep the target repo/dir in secrets. target-repository and target-directory are supplied from
  SYNC_DOCUMENTATION_TARGET_REPOSITORY and SYNC_DOCUMENTATION_TARGET_DIRECTORY (as before this PR), so the non-public target repo name and its internal layout are not hardcoded in the workflow. ci.yaml passes both secrets through to the real-sync job.
  - Restore reviewer routing via the action's new pr-reviewers input (dash0hq/operator,dash0hq/engineering-docs).
  - Fix the transformations-file path. It is resolved relative to source-root, so the repo-root .github/
  file needs ../../ (the previous value pointed at a nonexistent helm-chart/dash0-operator/.github/… path and would ENOENT).

  **Verification**

  Ran the action's transformation engine end-to-end against helm-chart/dash0-operator: all 12 pages (overview + 11 topics) are produced, exit code 0, the leading heading and badges are stripped, the values.yaml link is rewritten, docs/… links become extensionless siblings, and absolute URLs are untouched. The coverage guard correctly fails when a docs/*.md file has no files: entry.

  **Notes**

  - Navigation remains hand-maintained on dash0-website (no nav: block is emitted), matching prior
  behavior.
  - Secrets are retained, not removed: DASH0_DOCS_REPO_GITHUB_PAT, SYNC_DOCUMENTATION_TARGET_REPOSITORY,
  and SYNC_DOCUMENTATION_TARGET_DIRECTORY are still required for the real sync (they already exist in repo settings).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


basti1302:
Just cross-posting from the DM conversation here: Please run the workflow from this branch, and without the dry-run flag, so we can verify whether it still works correctly.

Currently I only see one manually triggered run off of `main`, and with the dry run flag.

image

.github/workflows/sync-docs/transformations.yaml
```diff
@@ -23,14 +23,14 @@
 #              description      The `description` rendered into the page's frontmatter.
 #              transformations  Optional list of transformations applied to this file only, after the common ones.
 #
-# Every *.md file in helm-chart/dash0-operator/docs (except dotfiles such as .docs-structure.md) must have a `files`
-# entry; otherwise apply-transformations.py fails. This guards against new topic files being silently dropped from the
-# sync.
+# `files` is the sole opt-in list: any *.md file in helm-chart/dash0-operator that is not declared here is ignored by
+# the sync. When adding a new topic file under docs/, add a matching `files` entry, otherwise the page is silently
+# skipped.
```
basti1302: This behavior is a downgrade in my opinion. Previosly, when we added a new page, the GH actions workflow would break, now it gets silently ignored.

It is only a matter of time until we add a new docs page and forget to add it here, and no one notices.

.github/workflows/sync-docs-to-website.yaml
```diff
@@ -6,22 +6,25 @@ name: Synchronize Helm chart docs to dash0.com/docs
...
-# release has been published. Additionally, ci.yaml invokes it with dry-run: true on every CI build), to verify early
+# release has been published. Additionally, ci.yaml invokes it with dry-run: true on every CI build, to verify early
 # that the transformations declared in sync-docs/transformations.yaml still apply cleanly. In dry-run mode, only the
-# transformation step runs; no PR is created for the website docs.
+# transformation step runs; no PR is created for the website docs and no token is required.
 #
-# The target repository and the target directory within it are configured via the repository secrets
-# SYNC_DOCUMENTATION_TARGET_REPOSITORY and SYNC_DOCUMENTATION_TARGET_DIRECTORY (the path of the directory holding the
-# documentation pages within that repository).
+# The sync targets the default documentation repository (dash0hq/dash0-website); the pages are written under
```
basti1302: Not sure if it matters but I would rather not leak internals, not even the repository name of a non-public repo. But maybe that's just me. :-)

That's why those were defined as secrets previously.

.github/workflows/sync-docs-to-website.yaml
```diff
@@ -62,89 +56,47 @@ jobs:
...
         run: |
-          target_path="target-repository/${TARGET_DIRECTORY}"
-          mkdir -p "$target_path"
-          # Copy the transformed tree (dash0-operator/overview.md and dash0-operator/*.md) into the target directory.
-          cp -R "${RUNNER_TEMP}/transformed-docs/." "$target_path/"
+          node .sync-docs-action/packages/transformation-engine/src/apply-transformations.ts \
+            helm-chart/dash0-operator \
+            helm-chart/dash0-operator/.github/workflows/sync-docs/transformations.yaml \
+            "${RUNNER_TEMP}/transformed-docs"
+          echo "This is a dry run: the doc transformations applied successfully, no pull request is created."
```
basti1302: This is pretty convoluted, we check out the gh action code, and then run it locally here. I think what we should do instead is to add a dry run flag to ash0hq/sync-docs, which copies applies all transformations but does not create a commit nor a PR.

basti1302: It also failed in the last CI run: https://github.com/dash0hq/dash0-operator/actions/runs/29488152608/job/87589399681?pr=1220

geertjanw: Done — that dry-run flag now exists in the action (sync-docs-action#6), so on 6458a08 both modes are single uses: with dry-run: ${{ inputs.dry-run }}; the check-out-and-run-by-hand block is gone. 
 
The failure was a transformations-file path resolved relative to source-root — fixed and CI is now green, the dry run as well as the real run on the branch: https://github.com/dash0hq/dash0-operator/actions/runs/29662116097

.github/workflows/sync-docs-to-website.yaml
geertjanw: Reverted on [6458a08](https://github.com/dash0hq/dash0-operator/commit/6458a0884c511df53fcc6ca5fd338ca0f8dd3340): target-repository/target-directory are read from the SYNC_DOCUMENTATION_TARGET_*secrets again (sync-docs-to-website.yaml:, passed through in ci.yaml), so the non-public repo name and layout are no longer in the workflow.

.github/workflows/sync-docs/transformations.yaml
geertjanw: Fixed on [6458a08](https://github.com/dash0hq/dash0-operator/commit/6458a0884c511df53fcc6ca5fd338ca0f8dd3340): added a coverage: guard so that a new page with no files: entry now fails the run (including the CI dry-run) instead of being silently skipped.

basti1302 (approved):
Looks good, thanks for improving this! 👍  A centrally managed GH action is for sure a much better way to manage this.

Should we merge the PR on the GH action, release it, and the directly pin a release of the action in this PR before merging it? I'll leave this up to you, just let me know.

geertjanw:
@basti1302: https://github.com/dash0hq/dash0-operator/pull/1220/changes/00c08d153026d233a03f8e0d0ed4a3c0a8693c92#diff-0300a4f8c9eb350461c2578293e8b46f84c45ba25c4f4866e2c6eb117a66c80a

mmanciop:
PR introduces no docs changes: https://github.com/dash0hq/dash0-operator/actions/runs/29837759924
Screenshot 2026-07-21 at 16 10 37

PR #1222, opened 2026-07-16
ci: remove claude code github workflow

basti1302:
It has been a while since anyone used "@claude review", this has been replaced largely by running AI-assisted reviews locally before marking the PR as ready for review. Looks like we no longer need this.


basti1302:
The context for this is also https://linear.app/dash0/issue/OPE-457/revisit-all-ci-actions-that-commit-and-push-collector-version-update, because theoretically the claude GH action can also commit and push (a feature we never used though).

p-d0 (approved):

PR #1223, opened 2026-07-16
ci: use gh to commit, so commits are automatically signed

basti1302:
For all automated workflows that commit and push, that is:
- update-collector-components.yaml
- update-opentelemetry-injector.yaml
- sync-go-version-with-dockerfile.yaml

For the record, I'm aware that we plan to replace sync-go-version-with-dockerfile.yaml with something else, but I don't want to break existing workflows when enforcing signed commits, hence this one gets updated in this PR as well.

### DIFF
--- a/.github/workflows/scripts/sync-go-version-with-dockerfile.sh
+++ b/.github/workflows/scripts/sync-go-version-with-dockerfile.sh
@@ -18,6 +18,11 @@ if [[ -z "${HEAD_REF:-}" ]]; then
   exit 1
 fi
 
+if ! command -v gh &> /dev/null; then
+  echo "Error: the gh executable is not available." >&2
+  exit 1
+fi
+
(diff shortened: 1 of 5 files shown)


basti1302:
## Before

Commits from github-actions bot were not signed:

image

image

## After

Commits from github-actions bot are now signed:

* https://github.com/dash0hq/dash0-operator/pull/1225
* https://github.com/dash0hq/dash0-operator/pull/1226

image
image

For the record, I did not test sync-go-version-with-dockerfile.yaml.

p-d0 (approved):
I wonder if that adds any security, though? If a token has been compromised, any commit made with `gh` and that token will show up as signed.

basti1302:
> I wonder if that adds any security, though? If a token has been compromised, any commit made with `gh` and that token will show up as signed.

Discussed offline. (This allows us to enforce a ruleset for all branches that requires signed commits.)

PR #1231, opened 2026-07-17
feat(collector): add k8s.node.uid to self-monitoring telemetry

basti1302:
The node UID is not available via the Kubernetes downward API and cannot be enriched by pipeline processors because self-monitoring telemetry is OTel-SDK-internal (it does not flow through the collector pipelines).

Add a custom internal-telemetry factory to the collector image build that wraps the default otelconftelemetry factory and, at collector startup, resolves the node UID via the Kubernetes API (using the already-injected K8S_NODE_NAME) and attaches it as k8s.node.uid to the self-monitoring resource. The lookup is best-effort: it never blocks or fails collector startup.

Also assert k8s.node.uid on the collector self-monitoring metrics and logs in the end-to-end tests.


basti1302:
Alternative approaches we might consider (as discussed on Monday):
* route telemetry from `service.telemetry` (e.g. the self-monitoring metrics and logs) through a k8s_attributes instance - not sure, if possible, needs research
* use an init container instead of a telemetry factory (I personally feel this is more heavy-weight than the approach in this PR, but it is obviously an alternative option)

This is currently work in progress and not ready to be reviewed, let alone merged.

basti1302:
I've looked at the alternatives we discussed a while back.

### Node UID Lookup in entrypoint script

**Pros:**
* no gke allowlist change
* no dependency on telemetry.Factory API

**Cons:**
* harder to test
* startup is serial, node uid lookup happens strictly before collector startup (in-process this can be done concurrently to some degree)
* Only covers the collector container, config-reloader and friends still need to do the lookup and add the attribute via the OTel Go SDK.
* Most importantly, this degrades badly. Just adding `k8s.node.uid: ${env:K8S_NODE_UID}` to the `service.telemetry.resource.attributes` will actually set `k8s.node.uid=""` or even `k8s.node.uid="<nil>"` if the env var is empty or absent. We would need the entrypoint script to rewrite the collector config to mitigate that. 

### Node UID Lookup in Init Container

E.g. let the init container do the lookup, write the uid to a shared emptydir volume, pick the value up in the collector entrypoint and export it there.

**Pros:**
* one lookup for all containers in the pod

**Cons:**
* requires maintaining an additional container image
* changes the pod spec, requires a new volume + mount 
* requires changes to the gke allowlist
* **the entrypoint script still would need to rewrite the configmap if the node uid is not available, see above**
* startup is serial, node uid lookup happens strictly before collector startup (in-process this can be done concurrently to some degree)
* delays every collector start additionally by a container launch

### The `telemetry.Factory` approach (this PR)

**Pros:**
* Lookup can happen concurrently in the collector process
* Degrades nicely (it just does not add the resource attribute if is not available)

**Cons:**
* Only covers the collector container, config-reloader and friends still need to do the lookup and add the attribute via the OTel Go SDK.

### route self-monitoring through k8s_attributes

There is no config-only way to do this, e.g. it is not possible to add `k8s_attributes` to  `service.telemetry` directly. The way this would work: point the periodic OTLP reader and the batch OTLP log processor at the collector's own receiver (or maybe a new dedicated pipeline with its own receiver for this), instead of pointing it directly at the main Dash0 export.  

**Pros:**
* No additional lookup code
* Enrichment happens automatically

**Cons:**
* Self-monitoring breaks when the pipeline it monitors breaks. E.g. let's say memory_limiter refuses data under pressure and drops it, this now also drops self-monitoring metrics and logs.
- Routing and graph startup failures go dark. For some failure modes the internal telemetry is currently independent of the main telemetry pipelines, so the logs explaining a failing exporter or extension still reach Dash0. When re-routing through the collector this will likely stop working.
- Log amplification. An export error produces a log record, which enters the pipeline, whose export fails, which produces another log record. This can create pretty nasty feedback loops. (I think this is the most problematic downside.)

### Resource Detectors for internal telemetry

https://github.com/open-telemetry/opentelemetry-collector/pull/15226 introduces resource detectors for internal telemetry.

Adds an experimental service::telemetry::resource::detection/development::detectors key, wired to go.opentelemetry.io/contrib/otelconf/x. But this only allows adding detectors from a very limited set of options: container, host, process, service.

This is not useful for our purposes.

## Summary

So far, the telemetry factory is still the best option IMO.

internal/collectors/otelcolresources/desired_state.go
```diff
@@ -502,6 +502,12 @@ func assembleDesiredState(
 			return desiredState, err
 		}
 		desiredState = append(desiredState, addCommonMetadata(signalControlCollectorConfigMap))
+		desiredState = append(
+			desiredState, addCommonMetadata(assembleServiceAccountForSignalControlCollector(config)))
+		desiredState = append(
+			desiredState, addCommonMetadata(assembleClusterRoleForSignalControlCollector(config)))
+		desiredState = append(
+			desiredState, addCommonMetadata(assembleClusterRoleBindingForSignalControlCollector(config)))
```
basti1302: _issue(blocking):_ as discussed, this does nothing as of today, since the sce collector image does not incorporate the telemetry factory. To be discussed how we want to resolve this.

basti1302: reverted

p-d0 (approved):

PR #1242, opened 2026-07-21
chore(claude): less noisy comments

basti1302:
Claude sometimes tends to elaborate on the history of humanity as a whole on one line code changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -26,6 +26,16 @@ The directory `test/util` contains additional Go code only used in unit tests.
 The directory `test/e2e` contains the end-to-end test suite.
 The directory `test-resources` contains a collection of scripts for running semi-manual tests scenarios.
 
+## Code Comments
+
+When making changes, never add comments regarding the state before your changes - code comments are not a place for
+history lessons.
+The motiviation for a specific change can be part of the commit comment (if you have been asked to commit).
+Add godoc comments for public functions.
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -26,6 +26,16 @@ The directory `test/util` contains additional Go code only used in unit tests.
 The directory `test/e2e` contains the end-to-end test suite.
 The directory `test-resources` contains a collection of scripts for running semi-manual tests scenarios.
 
+## Code Comments
+
+When making changes, never add comments regarding the state before your changes - code comments are not a place for
```
p-d0, line 31: when there was a good reason to not go with the simple/obvious solution, I think it makes sense to explain in a comment why it was changed, so a person (or bot these days) doesn't change it back to the undesired state

p-d0 (approved):

CLAUDE.md
basti1302: Fair. I find Claude far too often comments things like "this was previously implemented like this, now it is implemented like that" for no good reason whatsoever, and I need to go in and remove this boilerplate comments. Maybe that's a me problem related to my prompting style? I don't know.

I would hope that this part:
> You may use implementation comments to explain non-obvious aspects, but keep it as short as possible.

would cover the "not obvious" case to some degree, but maybe not.

@p-d0 Feel free to tweak CLAUDE.md further.

p-d0: approved because I have disabled all AI-project-specific settings on my end

PR #1248, opened 2026-07-22
feat(instrumentation): use Dash0 OpenTelemetry Python distribution

mmanciop:
Swap the Python auto-instrumentation payload from upstream `opentelemetry-distro` (plus hand-picked instrumentors and proto-http exporter) to [`dash0hq/opentelemetry-python-distribution`](https://github.com/dash0hq/opentelemetry-python-distribution), pinned by commit SHA.

The distribution vendors pure-Python OTLP/HTTP and OTLP/gRPC exporters (no `grpcio`, no `google.protobuf`) and pins the full upstream instrumentation set, so `requirements.txt` and `analyze-dependencies.sh` are gone, and `sitecustomize.py`'s `OTEL_EXPORTER_OTLP_PROTOCOL` gates collapse to a single `OTEL_EXPORTER_OTLP_ENDPOINT` → `DASH0_OTEL_COLLECTOR_BASE_URL` bridge.


images/instrumentation/python/sitecustomize.py
```diff
@@ -13,16 +13,13 @@
 minimum_python_minor_version = 10
 
 double_instrumentation_check_packages = [
+    "dash0-opentelemetry-distro",
```
basti1302: TBH I would have expected the sitecustomize script to become part of the distro

mmanciop: We could move it over indeed

images/instrumentation/python/sitecustomize.py
```diff
@@ -179,34 +176,16 @@ def import_distro():
...
-            "which prevented the Dash0 operator from setting its own values for OTEL_EXPORTER_OTLP_ENDPOINT/PROTOCOL, " +
-            "please remove OTEL_EXPORTER_OTLP_ENDPOINT and OTEL_EXPORTER_OTLP_PROTOCOL from the container if you want " +
-            "to use Python auto-instrumentation.)"
-        )
-        return
-    _log_debug("found eligible OTEL_EXPORTER_OTLP_PROTOCOL value: {}".format(otlp_protocol))
+    # The Dash0 Python distribution requires DASH0_OTEL_COLLECTOR_BASE_URL to activate and derives its OTLP endpoint
+    # from it. The Dash0 operator sets OTEL_EXPORTER_OTLP_ENDPOINT; bridge it here so the distribution activates
+    # without requiring operator-side changes.
+    if not os.environ.get("DASH0_OTEL_COLLECTOR_BASE_URL"):
```
p-d0: the operator does set `DASH0_OTEL_COLLECTOR_BASE_URL` here https://github.com/dash0hq/dash0-operator/blob/main/internal/workloads/workload_modifier.go#L643

mmanciop: Yep that guard is more for places where we may want to use the distro later. Anyhow, I am moving the sitecustomize in the distro, then update the PR, and we should be ready to ship it.

p-d0: > The Dash0 operator sets OTEL_EXPORTER_OTLP_ENDPOINT; bridge it here so the distribution activates without requiring operator-side changes.

ok, maybe I'm misreading this comment wrt the purpose of this `if` :)

p-d0 (approved):

PR #1255, opened 2026-07-24
feat(teams): add Dash0Team CRD and reconciler

mmanciop:
- Adds a new `Dash0Team` custom resource at group `dash0.com/v1alpha1` (deliberately distinct from `operator.dash0.com` — teams are a Dash0 platform object, not operator plumbing) with spec fields for display (name, description, color gradient) and members (strings, resolved server-side as either internal member IDs or emails).
- Adopts `dash0-api-client-go v1.17.0` (which carries the team envelope contract on the wire).


test/e2e/dash0-api-mock/main.go
```diff
@@ -76,6 +76,9 @@ func main() {
 	router.PUT("/api/notification-channels/:origin", handleNotificationChannelRequest)
 	router.DELETE("/api/notification-channels/:origin", handleNotificationChannelRequest)
 
+	router.PUT("/api/teams/:origin", handleTeamRequest)
```
p-d0, line 79: mock routes are registered but not used by any e2e tests (unless I missed something)

see https://github.com/dash0hq/dash0-operator/blob/main/test/e2e/e2e_test.go#L718 for notification channels as example

p-d0 (approved):

test/e2e/dash0-api-mock/main.go
mmanciop: Fixed in 27b4877

PR #1266, opened 2026-07-30
chore(agent0-connector): refine privileges

p-d0:
This PR flips the flag deny-list to an allow-list and treats configmaps as sensitive resources (similar to secrets).


basti1302 (approved):
👍 👍

PR #1267, opened 2026-07-30
feat: add Ruby auto-instrumentation

mwear:
Adds Ruby to the runtimes the operator can auto-instrument, using the [Dash0 OpenTelemetry Ruby distribution](https://github.com/dash0hq/opentelemetry-ruby-distribution). Opt-in via `operator.instrumentation.enableRubyAutoInstrumentation` (default `false`), mirroring Python.

### Summary

- **Instrumentation image**: two build stages (glibc/musl) installing the distribution and its
  pinned dependency closure into a plain `GEM_HOME` with the entry symlink the injector requires.
  The distribution is not published to RubyGems, so its source tree is fetched at a pinned git
  ref and built during the image build, mirroring the Python stages.
- **Injector configs**: adds `injector-with-ruby.conf` and `injector-with-python-and-ruby.conf`,
  so every combination of the Python and Ruby opt-ins has a config file.
- **Operator**: the Helm setting becomes `DASH0_ENABLE_RUBY_AUTO_INSTRUMENTATION`, which selects
  the injector config file for instrumented containers. Selection is centralised in
  `otelInjectorConfigFileFor`, used both when writing `OTEL_INJECTOR_CONFIG_FILE` and when
  deciding whether existing workloads need updating.
- **Tests**: a Ruby runtime in the instrumentation-image suite (6 cases across 4 base images,
  covering glibc and musl, the Ruby 3.3 minimum, and an unsupported 3.2 that must stand down),
  plus a Rails application and the end-to-end wiring for it.
- **Docs**: supported runtimes, the opt-in, the Ruby 3.3 minimum, the `http/protobuf`
  requirement, fail-open behaviour, and `DISALLOWED_LIB_PATH`.

### Verification

- End-to-end on kind: `should instrument new Ruby deployments` and `should instrument and
  uninstrument all workload types` both pass — webhook instrumentation through to matching spans
  from a Bundler-managed Rails application.
- Instrumentation-image tests: 24/24 for Ruby on both amd64 and arm64; Python and Node unaffected
  (78/78).
- The multi-arch image build produces correct bundles for all four arch × libc combinations.

### Notes

- Default installations are unchanged apart from the image carrying the bundle; Ruby is not
  instrumented unless the setting is enabled.
- The number of injector config files doubles with each opt-in runtime. A follow-up can collapse
  them into a single file plus a per-container `OTEL_INJECTOR_AUTO_INSTRUMENTATION_DISABLED`,
  which the injector honours as a complete override of the file's list.
- The distribution is pinned by commit SHA; this can move to a release tag once one is cut.


mwear:

mwear:
closing. will reopen from a non-fork branch.

PR #1269, opened 2026-07-31
feat(target-allocator): add support for allowInsecureAuthSecrets flag

p-d0:


helm-chart/dash0-operator/docs/metrics-and-scraping.md
```diff
@@ -151,6 +153,19 @@ operator:
       clientCertSecretName: "ta-mtls-client-cert-secret"
 ```
 
+#### Allowing Insecure Auth Secrets
+
+If the transport security between the OpenTelemetry Target Allocator and the collectors is already guaranteed by other means, for example by a service mesh, you can skip the mTLS setup and let the Target Allocator serve the credentials over plain HTTP instead:
```
mmanciop, line 158: I'd rather leave a not with "Don't use this unless you have other means of securing..."

test-resources/bin/test-scenario-06-ta-mtls.sh
```diff
@@ -24,7 +24,8 @@ runtime_under_test="nodejs" # nodejs is currently the only runtime with a /metri
 additional_namespaces="false"
 USE_CERT_MANAGER="true"
 PROMETHEUS_CRD_SUPPORT_ENABLED="true"
-TA_MTLS_ENABLED="true"
+TA_MTLS_ENABLED="${TA_MTLS_ENABLED:-true}"
```
mmanciop, line 27: What happens if both TA_MTLS_ENABLED and TA_ALLOW_INSECURE_AUTH_SECRETS are set to true? TA_ALLOW_INSECURE_AUTH_SECRETS skips CA check?

p-d0: my understanding is that they don't interfere (but it's not very useful to use both except for temporary testing maybe)

`TA_MTLS_ENABLED`: expose scrapeconfig over https, requiring a client certificate
`TA_ALLOW_INSECURE_AUTH_SECRETS`: will serve the full/unredacted scrapeconfig over http (normally the credentials are redacted when calling the http endpoint)

helm-chart/dash0-operator/docs/metrics-and-scraping.md
p-d0: Isn't that what "Only enable this setting if transport-level security is guaranteed by other means,..." already says?

test-resources/bin/test-scenario-06-ta-mtls.sh
p-d0: To clarify: `TA_ALLOW_INSECURE_AUTH_SECRETS` has no effect on the https connection. It basically controls whether the token/secrets are redacted or not when calling the http endpoint. By default (without the flag), sensitive values are replaced with `<secret>` in the scrapeconfig. By setting `TA_ALLOW_INSECURE_AUTH_SECRETS` to `true`, the actual values will be used.

helm-chart/dash0-operator/docs/metrics-and-scraping.md
mmanciop: Not as clearly

mmanciop (approved):

PR #1297, opened 2026-08-13
chore(agent0-connector): redact secrets from Dash0 resources

basti1302:
Redact secrets like Dash0 auth tokens or third-party credentials
from the responses to requests like
`kubectl get dash0operatorconfigurations -o yaml` before they
leave the cluster. Reject requests for Dash0 CRs when the output
cannot be redacted.

Secrets are redacted by parsing the response, then replacing the
potential secrets (e.g. export tokens etc.), then rendering the
document again. Re-rendering reproduces kubectl's output byte for byte,
because kubectl serializes a custom resource from the same unstructured
form.

Requests targeting resources with potential secrets are only allowed
with formats that can be parsed and interpreted, e.g. `-o json` and
`-o yaml`. Formats that reshape the response
(`-o go-template/jsonpath/custom-columns`, `--template`) as well
as `kubectl describe` are rejected.

Independently of redaction, the output-format allowlist now also
rejects the formats that render a local file (`go-template-file`,
`jsonpath-file`, ...), which could otherwise allow reading arbitrary
files in the connector's container, such as its service account
token.

`api/operator/credential_field_coverage_test.go` adds a heuristical
safeguard to catch missing redaction when Dash0 CRDs are added or
changed.

This also breaks up the monolithic images/agent0-connector/src into
multiple packages, which had become necessary to keep the added
logic maintainable.


p-d0 (approved):
Looks good 👍 Some more general comments

- `Dash0SyntheticCheck` can technically also carry sensitive data like tokens in the body
- `OTEL_EXPORTER_OTLP_HEADERS` should also be redacted unless we remove all env vars anyway

basti1302:
> Looks good 👍 Some more general comments
> 
> * `Dash0SyntheticCheck` can technically also carry sensitive data like tokens in the body
> * `OTEL_EXPORTER_OTLP_HEADERS` should also be redacted unless we remove all env vars anyway

I was a bit on the fence about the synthetic check's body, but you are right. I'll add it to the redaction.

Env var redaction is deliberately out of scope for this PR (it is already big), and yes, the current decision is to redact all env vars values on all workloads (tracked in OPE-520).

PR #1300, opened 2026-08-14
fix(collectors): extract k8s.node.uid via k8sattributes

mmanciop:
`k8s.pod.*` / `k8s.container.*` metrics from the k8s_cluster receiver, and Pod-kind Kubernetes events normalized by transform/k8s_events, only ever carried the node by name (a plain string), never by uid. Anything that resolves node identity by uid saw these as a different node than the one described by `k8s.node.*` metrics, which do carry `k8s.node.uid`.

Adding `k8s.node.uid` to the k8sattributes metadata list lets it resolve the uid via the pod association it already performs, reusing the Node informer it already maintains.

### DIFF
--- a/internal/collectors/otelcolresources/daemonset.config.yaml.template
+++ b/internal/collectors/otelcolresources/daemonset.config.yaml.template
@@ -784,6 +784,7 @@ processors:
       - k8s.job.uid
       - k8s.namespace.name
       - k8s.node.name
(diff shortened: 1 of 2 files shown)


p-d0 (approved):
as discussed, the addition in the daemonset is most likely redundant (it's already set by the `resourcedetection` processor) but we can add it for consistency

PR #1301, opened 2026-08-17
feat(sce): move sce components into separate collector

p-d0:
Moves the SCE components from the standard daemonset collector into their own collector (deployment).

Zone-awareness is implemented as "best-effort":
- if there is a collector in the same AZ the telemetry will be routed there
- if there's fewer replicas than zones, telemetry from AZs that don't have a collector replica will be routed to a collector from another AZ to not drop data or block the deployment
- in this case a warning will be logged, advising the user to increase the number of replicas to at least the number of AZs

Out-of-scope (upcoming work):

- SCE-only mode (i.e. do not deploy the standard collector at all)


helm-chart/dash0-operator/values.yaml
```diff
@@ -921,6 +921,116 @@ operator:
         timeoutSeconds: 1
         failureThreshold: 45
 
+    # The number of replicas of the Signal Control collector Deployment managed by the operator. This collector only
+    # exists when Signal Control is enabled (see operator.signalControl.enabled); it receives the Dash0-bound telemetry
+    # of the other two collectors, applies Signal Control and exports to Dash0. More replicas spread the tail-sampling
+    # reservoir and the RED-metrics cardinality tracking over more processes; fewer replicas use less memory.
+    signalControlCollectorReplicas: 2
```
basti1302, line 928: q: This PR adds new settings under `operator.collectors`, all of them prefixed with `signalControl`. Should we move them to `operator.signalControl` or `operator.signalControl.collector` instead?

internal/collectors/otelcolresources/desired_state.go
```diff
@@ -1815,6 +1847,335 @@ func assembleDeploymentCollectorContainer(
 	return collectorContainer, nil
 }
 
+// assembleSignalControlCollectorService exposes the Signal Control collector deployment to the other collector
```
basti1302, line 1850: q: Should assembling the sce collector related resources go somewhere under the `signalcontrol` package, instead of being here in `collector/otelcolresources`? I guess both has pros and cons, and they fit into both categories. I don't have a strong opinion either way, but I'm curious what your thoughts are about that.

internal/collectors/otelcolresources/exporters.partial.yaml.template
```diff
@@ -0,0 +1,102 @@
+{{- /*
+The exporters section is identical for all collector workloads: every workload declares every configured exporter
+(default and per-namespace). Which of them a workload actually references depends on its pipelines. This partial is
+parsed into all three collector configuration templates, see collector_config_maps.go.
```
basti1302, line 4: 👍  Nice! We could maybe extract more shared templates like this in the future, I assume we have a few more blocks that are somewhat repetitive.

helm-chart/dash0-operator/docs/advanced-configuration.md
```diff
@@ -431,8 +465,9 @@ operator:
 ## Configuring Pod-Level sysctls for the Collector Pods (TCP Keepalive)
 
 Pod-level [sysctls](https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/) can be applied to the
-collector pods via `operator.collectors.daemonSetSysctls` (for the daemonset collector) and
-`operator.collectors.deploymentSysctls` (for the cluster-metrics-collector deployment).
+collector pods via `operator.collectors.daemonSetSysctls` (for the daemonset collector),
+`operator.collectors.deploymentSysctls` (for the cluster-metrics-collector deployment) and
+`operator.collectors.signalControlCollectorSysctls` (for the SignalControl collector deployment).
 Both default to being unset, so no sysctls are applied unless you configure them.
```
basti1302:
```suggestion
All three default to being unset, so no sysctls are applied unless you configure them.
```

internal/util/cluster/kubernetes_version.go
```diff
@@ -27,6 +27,15 @@ const (
...
 	imageVolumesAlwaysMinimumMinorVersion = 31
+
+	// trafficDistributionMinimumMajorVersion is the lowest Kubernetes major version that knows the service field
+	// spec.trafficDistribution. Used together with trafficDistributionMinimumMinorVersion.
+	trafficDistributionMinimumMajorVersion = 1
+
+	// trafficDistributionMinimumMinorVersion is the lowest 1.x Kubernetes minor version that knows the service field
+	// spec.trafficDistribution (1.30, alpha; beta in 1.31, GA in 1.33). Used together with
+	// trafficDistributionMinimumMajorVersion.
+	trafficDistributionMinimumMinorVersion = 30
```
basti1302: q: Should we require 1.31 for this as well? AFAICT it is behind a [feature flag](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-graduated-or-deprecated-features) in 1.30, so it might not be enabled in 1.30 clusters.

helm-chart/dash0-operator/values.yaml
basti1302, line 928: q: Should we add validation for signalControlCollectorReplicas > 0 when signal control is actually enabled, e.g. enabled via Helm and enabled via a signal control CR?

OTOH, it is probably very unlikely that customers set this to 0 and also try to enable it, so maybe such a validation would be overkill.

internal/collectors/otelcolresources/desired_state.go
```diff
@@ -1815,6 +1847,335 @@ func assembleDeploymentCollectorContainer(
...
+			collectorResources.Requests = requests
+		}
+	}
+
+	collectorEnv, err := assembleCollectorEnvVars(config, workloadNameEnvVar, resourceRequirements.GoMemLimit, true)
+	if err != nil {
+		return corev1.Container{}, err
+	}
+
+	// (8) No --feature-gates argument: this collector runs no resourcedetection processor, all resource detection has
```
basti1302: `(8)`?

internal/collectors/collector_manager.go
```diff
@@ -198,6 +215,72 @@ func (m *CollectorManager) ReconcileOpenTelemetryCollector(
...
+	logger logd.Logger,
+) {
+	if m.clientset == nil {
+		return
+	}
+	// The clientset bypasses the controller-runtime cache on purpose: reading nodes through the cached client would
+	// start an informer that keeps every node object in memory for the lifetime of the operator. The read is served
+	// from the API server's watch cache (ResourceVersion "0") and restricted to nodes that carry a zone label, since
+	// nodes without one contribute nothing to the zone count.
+	nodes, err := m.clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{
```
basti1302: possible optimization: we could use a metadata client to only fetch the metadata (looks like we only read the labels). Might be good to not deserialize the whole node (especially node status) on every reconcile.

internal/collectors/otelcolresources/collector_config_maps_test.go
basti1302, line 2140: issue: I think for these tests to become valid, a non-empty SignalControlCollectorImage in the oTelColConfig has to be provided, due to the way how `func (c *oTelColConfig) signalControlGatewayActive()` is implemented. It looks like currently even for sc-full, signalControlGatewayActive returns false in this test, and then it renders a config with `$gwActive = false`.

basti1302 (approved):

helm-chart/dash0-operator/values.yaml
p-d0: `# Settings related to the collectors managed by the operator.`

Personally, I think it fits where it is.

internal/util/cluster/kubernetes_version.go
p-d0: But then it won't work in 1.30 even if the feature flag is enabled (unless I misinterpret what you are suggesting).

helm-chart/dash0-operator/values.yaml
p-d0: If the user sets 0, it will use the default 2 instead. We could add a validation but it would divert from similar fields like `operator.replicaCount` and `operator.signalControl.edgeProxy.replicas` which also don't have a validation in the Helm chart.

internal/util/cluster/kubernetes_version.go
p-d0: I've updated it to 1.31 to choose the more robust option. If we ever have a user that runs 1.30, can't upgrade, and needs that option, we can address it.

helm-chart/dash0-operator/values.yaml
basti1302: > If the user sets 0, it will use the default 2 instead.

I wasn't aware of that/didn't think of that.

internal/collectors/otelcolresources/desired_state.go
p-d0: I agree it could live in both packages. It does currently share the `oTelColConfig` and the template partial that live in `otelcolresources` - I'm not saying that this could not be refactored to move it elsewhere, but since it is a collector, it might just as well live in `collector/otelcolresources` (imo).

internal/collectors/collector_manager.go
p-d0: I've updated it, but I think a bigger thing to optimize might actually be the frequency of the requests - the number of AZs doesn't change very frequently so checking once an hour might be good enough for a warning log :thinking:

basti1302: Yes, absolutely. 👍

PR #1362, opened 2026-09-02
chore: add lint check for missing Ginkgo suite bootstrap files

basti1302:
This adds a make lint step that checks whether every directory that contains Ginkgo test also has a suite bootstrap file. Without such a bootstrap, `go test` compiles the specs and reports the package as passing, but the specs are not executed.

This is motivated by internal/util/zap actually missing such a suite bootstrap, the tests in it never ran. Production code had also diverted from the original tests, which were broken by now (and are fixed now).


basti1302:
Note: This came up incidentally while investigating a potential memory leak in `internal/util/zap/delegating_zap_core.go`. There will probably be a follow up PR to fix that leak.

p-d0 (approved):

PR #1364, opened 2026-09-02
chore(sc): check number of AZs less frequently

p-d0:
## Description

Since the number of AZs doesn't change very often, this PR updates the implementation to only refresh the cached number every 30 minutes. This doesn't have any functional consequences, it is only used for logging a warning when there are fewer SC collectors than AZs (or a short info when the issue is resolved).

## Authorship

### DIFF
--- a/internal/signalcontrol/scresources/sc_resources.go
+++ b/internal/signalcontrol/scresources/sc_resources.go
@@ -154,6 +154,9 @@ func (m *SignalControlResourceManager) updateResource(
 	if err = patch.DefaultAnnotator.SetLastAppliedAnnotation(desiredResource); err != nil {
 		return false, err
 	}
+	// Carry over the live resource version; some kinds (PodDisruptionBudget) reject an update without it
+	// ("resourceVersion: Invalid value: 0: must be specified for an update") and it adds optimistic locking.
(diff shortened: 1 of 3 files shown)


internal/signalcontrol/scresources/sc_resources.go
```diff
@@ -154,6 +154,9 @@ func (m *SignalControlResourceManager) updateResource(
 	if err = patch.DefaultAnnotator.SetLastAppliedAnnotation(desiredResource); err != nil {
 		return false, err
 	}
+	// Carry over the live resource version; some kinds (PodDisruptionBudget) reject an update without it
+	// ("resourceVersion: Invalid value: 0: must be specified for an update") and it adds optimistic locking.
+	desiredResource.SetResourceVersion(existingResource.GetResourceVersion())
```
p-d0, line 159: It might also make sense to set it on the `agent0-connector` (despite not using a PDB atm), but since it doesn't cause any errors there, I didn't want to update unrelated resources in this PR.

basti1302: Sounds good.

basti1302 (approved):

PR #1369, opened 2026-09-03
fix: fix memory leak in DelegatingZapCore

basti1302:
DelegatingZapCore appended each core derived via `With` to a slice named `dc.clones` on the parent, so that SetDelegate and UnsetDelegate could reach cores derived before the call. The root core lives for the lifetime of the process, and every logger ever derived stayed reachable from it via `clones`, together with its concatenated field slice and its otelzap.Core clone holding the converted OTel attributes.

The heap grew with every webhook request, and never shrank. Once the live heap reaches GOMEMLIMIT, GC runs back to back and eats the container's CPU budget.

This inverts the reference direction: the delegate now lives in a sharedDelegate (singleton) that the root core and all derived cores point to. A derived core resolves its own delegate lazily on first use and memoizes it. Nothing points from a parent to a derived core any more, so a derived core is eligible for garbage collection as soon as the logger holding it goes out of scope.

### DIFF
--- a/internal/util/zap/delegating_zap_core.go
+++ b/internal/util/zap/delegating_zap_core.go
@@ -10,17 +10,82 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
-// DelegatingZapCore is a Zap core that keeps messages in a size-limited buffer until another core is added as a
-// delegate via SetDelegate. After that, the core will pass all messages to the delegate. The messages that are already
-// buffered can then be spooled to the new core via DelegatingZapCoreWrapper.LogMessageBuffer.ForAllAndClear.
+// DelegatingZapCore is a Zap core that keeps messages in a size-limited buffer until an actual zapcore.Core is added as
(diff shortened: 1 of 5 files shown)


internal/util/zap/delegating_zap_core.go
```diff
@@ -10,17 +10,82 @@ import (
...
+// additional fields (which will be included in every log record written by that child logger). There are no references
+// from parent to child or vice versa.
 type DelegatingZapCore struct {
-	delegate         atomic.Pointer[zapcore.Core]
+	sharedDelegate   *sharedDelegate
+	derivedDelegate  atomic.Pointer[derivedDelegate]
 	logMessageBuffer *Mru[*ZapEntryWithFields]
 	level            zapcore.Level
 	fields           []zapcore.Field
-	clones           []*DelegatingZapCore
```
basti1302, line 21: ^ tracking the list of child loggers (here called clones) in the parent was the memory leak

p-d0 (approved):

PR #1379, opened 2026-09-09
chore(self-monitoring): treat the collector container separately

basti1302:
Self-monitoring for the collector pods was enabled by applying the same
set of environment variables to every container of the pod spec. That
treats two fundamentally different mechanisms as one: the sidecars
(configuration-reloader, filelog-offset-sync) run the OpenTelemetry Go
SDK, which is configured exclusively via environment variables. In
contrast, the opentelemetry-collector container configures its internal
telemetry via the service::telemetry section of the collector configuration.

The previous logic set five environment variables that the collector
container does not read: OTEL_EXPORTER_OTLP_ENDPOINT, _PROTOCOL and
_HEADERS, and OTEL_LOG_LEVEL. For gRPC/HTTP exports with secret-backed
headers it also received a second copy of those secrets as environment
variables, under the SELF_MONITORING-suffixed names, while the rendered
self-monitoring pipeline references the DEFAULT_0-suffixed ones
injected by the exporters section.

The only thing the collector container actually needs is
SELF_MONITORING_AUTH_TOKEN, which the pipeline refers to via
${env:SELF_MONITORING_AUTH_TOKEN}.

OTEL_RESOURCE_ATTRIBUTES was actually honored by the internal telemetry
in principle, but only partly, because most of the service.* attributes
were then overwritten by fixed attributes baked into the collector
build (service.name, service.version). Only the service.namespace would
be taken from the OTEL_RESOURCE_ATTRIBUTES, this is now set explicitly
via the config maps instead.

The previous behavior was a potential future foot-gun: If a future
collector version starts honoring OTEL_EXPORTER_OTLP_ENDPOINT for its
internal telemetry, the collector would silently gain a second exporter
and double-send self-monitoring telemetry after a routine version bump.

EnableSelfMonitoringInDaemonSet/-InDeployment now take the name of the
container running the collector, and treats that container differently.
That container only gets the auth token, all other containers get the
full set of Go SDK environment variables.


basti1302:
> ### ❌ E2E tests failed
> * **Branch**: `treat-collector-containers-separately-when-enabling-self-monitoring`
> * **Run**: [#34579084911](https://github.com/dash0hq/dash0-operator/actions/runs/34579084911)

Looks like this refactoring broke attaching a service.namespace.

basti1302:
> Looks like this refactoring broke attaching a service.namespace.

Fixed.

p-d0 (approved):

PR #1383, opened 2026-09-10
feat: make the operator compatible with OpenShift

p-d0:
## Description

The following adaptions have been made to support installing the operator in OpenShift via Helm:

- added a custom SCC for the daemonset collector
- dropped the `runAsUser/runAsGroup` when running on OpenShift, so either the user id from the image is used or one is generated based on the configured range
- small change to RBAC since OpenShift/MicroShift enables the [OwnerReferencesPermissionEnforcement](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement) admission plugin (not gated behind the OpenShift flag since other clusters could technically enable it as well, even though it's off by default)

## Authorship

### DIFF
--- a/images/instrumentation/Dockerfile
+++ b/images/instrumentation/Dockerfile
@@ -155,4 +155,7 @@ COPY --from=build-ruby-glibc /dash0-instrumentation/ruby /dash0-instrumentation/
 COPY --from=build-ruby-musl /dash0-instrumentation/ruby /dash0-instrumentation/agents/ruby/musl
 
 WORKDIR /
+# A numeric non-root USER lets the kubelet verify runAsNonRoot when the pod-level runAsUser is dropped. Matches
+# defaultInitContainerUser.
+USER 13020
 CMD ["/copy-instrumentation.sh"]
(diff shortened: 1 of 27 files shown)


images/instrumentation/Dockerfile
```diff
@@ -155,4 +155,7 @@ COPY --from=build-ruby-glibc /dash0-instrumentation/ruby /dash0-instrumentation/
 COPY --from=build-ruby-musl /dash0-instrumentation/ruby /dash0-instrumentation/agents/ruby/musl
 
 WORKDIR /
+# A numeric non-root USER lets the kubelet verify runAsNonRoot when the pod-level runAsUser is dropped (OpenShift,
+# where the namespace SecurityContextConstraints assigns an in-range UID instead). Matches defaultInitContainerUser.
+USER 13020
```
basti1302, line 161: _question:_ Should we also bake in a group ID? Or alternatively, maybe add a sentence to the comment why a baked-in group id is not desirable?

basti1302 (approved):
👍

images/instrumentation/Dockerfile
p-d0: Thanks, added to the comment that the GID is not relevant for `runAsNonRoot` so it is not set.

PR #1386, opened 2026-09-11
ci(gke-ap): fix failing gke autopilot check

basti1302:
Only wait for at least one daemonset collector pod to get ready instead of waiting for the entire daemonset to become ready via "kubectl rollout status".

When the GKE AutoPilot cluster scales up (adds more nodes), the DESIRED goal for the daemonset increases with each node. For each new node it takes time until the node is ready for scheduling. Whether the entire daemonset becomes ready within a given timeout is quite non-deterministic and depends entirely on GKE AutoPilot auto-scaling behavior, not on whether the allowlist matches the daemonset.

We are only interested in whether the allowlist matches and collector pods generally are able to start.

One collector daemonset pod is good enough to prove that the matched allowlist permits the workload to actually run.

Also:
* increase CI timeout for GKE AutoPilot AllowList check
* add timestamps to gke-ap-workload-allowlist-check.sh output
* do not double-wrap diagnostics dir (tar.gz in zip)
* add node information to diagnostics

Disclaimer: It is not entirely clear, why the check started failing, but the behavior that made it fail more often recently seems to be a change in the GKE AutoPilot auto-scaling behavior or potentially a different node type. Either way, the individual pods of the daemonsets got ready just fine, the daemonset just never got ready in the allotted 90s because it was chasing up to 10 nodes coming up every few seconds.

### DIFF
--- a/.github/workflows/gke-ap-workload-allowlist-check.yaml
+++ b/.github/workflows/gke-ap-workload-allowlist-check.yaml
@@ -121,7 +121,7 @@ jobs:
       always() &&
       (needs.build_and_push_images.result == 'success' || needs.build_and_push_images.result == 'skipped') &&
       (github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success')
-    timeout-minutes: 20
+    timeout-minutes: 15
     permissions:
       contents: read
       pull-requests: write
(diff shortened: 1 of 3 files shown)


p-d0 (approved):
👍

PR #1394, opened 2026-09-15
chore(collectors): replace deprecated OTel component aliases in collector configs

thieltges:
## Description

The collector logs deprecation warnings at startup for four component aliases that collector-contrib
renamed in [v0.153.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.153.0)
and later. This renames them in the collector config templates.

| deprecated | current | |
|---|---|---|
| `resourcedetection` | `resource_detection` | processor |
| `hostmetrics` | `host_metrics` | receiver |
| `kubeletstats` | `kubelet_stats` | receiver |
| `k8snode` | `k8s_api` | a resource detection *detector*, not a component |

Checked against the pinned collector version (v0.160.0) before renaming: each component's
`metadata.yaml` carries the new name as `type` and the old one as `deprecated_type`. For the
detector, `internal/k8sapi/config.go` maps both `mapstructure:"k8s_api"` and `mapstructure:"k8snode"`
onto the same config struct, so `node_from_env_var` still binds and the emitted attributes come from
the same code path.

Closes OPE-366. Unblocked by OPE-255, which lifted the version pin on the resource detection processor.

### Compatibility

The aliases work in one direction only. Old names keep resolving in new collectors, which is why this
is safe to merge, but new names do **not** resolve in collectors older than 0.153.0. Anyone who pins
`operator.collectorImage.tag` or `.digest` to a pre-0.153.0 image while upgrading the chart will get
`unknown type: "kubelet_stats"` and a collector that never starts, rather than today's deprecation
warning. The chart ships a matched collector by default, so this only affects explicitly pinned images
— worth a line in the upgrade notes.

### What is deliberately not renamed

These strings look like the same rename but are not:

- `operator.collectors.kubeletstats.*` in `values.yaml` — a Helm value, i.e. chart API. Renaming it
  would break every user's values file.
- `--feature-gates=-processor.resourcedetection.propagateerrors` — a feature gate ID. Gate IDs are not
  aliased.
- gomod paths in `images/collector/src/builder/config.yaml` and doc links to `kubeletstatsreceiver`
  etc. — upstream Go module and directory names, unchanged upstream.

User-facing docs in `helm-chart/dash0-operator/docs/platform-specific.md` still say `kubeletstats` and
`hostmetrics`. Left as-is deliberately: those lines describe the feature, and renaming them there
reads as if the Helm value changed too.

## Verification

- `make go-unit-tests`, `make helm-unit-tests`, `make golangci-lint` all pass.
- Deployed to a kind cluster via `test-scenario-02-operator-cr-aum.sh`. The deprecation warnings are
  gone from every collector pod, no errors replaced them, and both daemonset and deployment pods reach
  ready. The live config maps carry `kubelet_stats`, `host_metrics`, `resource_detection` and the
  `k8s_api` detector, so the collector genuinely loaded the new names rather than the templates merely
  rendering them.

## Authorship

- [ ] personally wrote the PR description.


thieltges:
@claude review

thieltges:

basti1302 (approved):
👍 LGTM 👍

PR #1395, opened 2026-09-15
chore: support configuring the operator with env vars instead of args

p-d0:
## Description

- allows configuring the operator purely with env vars (if both args and env vars are set, args win)
- the Helm chart continues to use the args normally, this PR is mainly preparation work for supporting OLM
- adds a temporary script for testing - the script can be removed once OLM support has been added (then the env vars will be tested via the OLM tests)

## Authorship


test-resources/bin/test-env-var-fallback.sh
```diff
@@ -0,0 +1,197 @@
+#!/usr/bin/env bash
```
p-d0, line 1: as mentioned in the description, this script is a temporary way of testing the env fallback - will be removed once OLM support is implemented

basti1302 (approved):
👍 

I personally still think that just removing all command line flags from operator_manager_startup and from the Helm chart, and converting them all to env vars would be slightly better. But supporting both is of course also feasible. 👍

p-d0:
> 👍
> 
> I personally still think that just removing all command line flags from operator_manager_startup and from the Helm chart, and converting them all to env vars would be slightly better. But supporting both is of course also feasible. 👍

I never said, we can/should not remove them at some point. But there's already enough feature-creep wrt OpenShift support and changing a rather fundamental part of the operator that affects all users can be implemented independently in the future.

Most software supports multiple levels of configuration (env vars, args, config file). I don't think having the env vars in addition to the args is an anti-pattern.

basti1302:
> Most software supports multiple levels of configuration (env vars, args, config file). I don't think having the env vars in addition to the args is an anti-pattern.

That concern is immaterial here though, since nobody starts or configures the executable directly. It is always only configured via the Helm chart.

> I never said, we can/should not remove them at some point. But there's already enough feature-creep wrt OpenShift support and changing a rather fundamental part of the operator that affects all users can be implemented independently in the future.

That is a good point. Doing it like this as an intermediate step makes sense, and reduces the risk. 👍

p-d0:
> That concern is immaterial here though, since nobody starts or configures the executable directly. It is always only configured via the Helm chart.

It wasn't intended as a concern. Rather to emphasize that I don't see the reasoning behind having such a strong aversion to supporting both env vars and args for the time being. Just dropping support for args from one release to the next seems like an unnecessary step at this point.

Additionally, the args and env vars are not maintained separately in the code, so there is no possible drift or maintenance overhead.

PR #1401, opened 2026-09-16
docs: correct Node.js minimum version to 18.19+

EoinTrial:
Auto-instrumentation's Node.js requirement was listed as 16+, but the distribution it points to (`opentelemetry-js-distribution`) is on OpenTelemetry JS 2.x, which dropped Node 14/16 support.

This PR updates the references to 18.19+ in both the `README.md` and `helm-chart/dash0-operator/README.md`.


basti1302:
Hi Eoin,

our Node.js distribution comes with both OpenTelemetry 1.x and 2.x and automatically [switches](https://github.com/dash0hq/opentelemetry-js-distribution/blob/30ccd86abdfe02635779f76a04337da4ce569e46/src/index.ts#L7-L13), based on the Node.js version. There are automated tests for all supported versions, including Node.js 16.x. While we might decide to drop support for 16.x at some point, the current state of the docs is correct.

image

PR #1405, opened 2026-09-18
chore(agent0-connector): tighter validation and redaction

basti1302:
This is another pass at making agent0-connector more secure by default. While the previous iterations already focussed heavily on making the component secure, this PR makes the restrictions for executing `kubectl` even tighter. Once more, we trade a bit of power of the operator-agent0 integration for increased security by default. That is, Agent0 (or a hypothetical malicious actor abusing the feature) now can do a less things over this integration, but it in exchange it has tighter security constraints.

Changes:
* Redact every resource type: Previously, redaction was bound to a hardcoded selection of resource types, everything else was handed to the backend as kubectl rendered it. Now redaction is applied to all `kubectl get` command, independent of the resource type. Redaction for resource types that are not in the default RBAC rule is by its very nature a best effort.
* Reject kubectl describe for all resource types: Since we now redact all resource types, only rejecting kubectl describe for a specific set of resource types is no longer reasonable. 
* Redact the *command line* of workload containers, in addition to its environment variables, redact command and args. Also redact the command of an exec probe or lifecycle hook. Although potentially less common, the command line arguments can also contain secrets, e.g. an API key passed as "--api-key=..." etc.
* Keep JSON numbers intact across the redaction round trip: Fix how JSON numbers are parsed and re-rendered after redaction.
* Redact parseable annotations also in nested structures, and both for JSON-shaped and YAML-shaped annotations.
* Build the kubectl environment from an allowlist, instead of letting it inherit the full environment. Specifically exclude DASH0_AGENT0_CONNECTOR_AUTH_TOKEN and the self-monitoring exporter headers, which kubectl does not need.
* Redact potential credential values that are scalars, but not strings (e.g. { "api-token": 123456 }).
* Apply the --sort-by guard to "kubectl get" only, it is irrelevant for other kubectl subcommands.
* Recover from panics in the command executor. 
* Improve wording in values.yaml to make it even clearer that using custom RBAC rules shifts some responsibility to the user.

## Authorship


api/operator/credential_field_coverage_test.go
```diff
@@ -111,106 +104,10 @@ func TestAgent0ConnectorRedactsEveryCredentialField(t *testing.T) {
...
-		}
-		accepted := []string{strings.ToLower(names.Kind)}
-		for _, name := range append([]string{names.Singular, names.Plural}, names.ShortNames...) {
-			if name != "" && !slices.Contains(accepted, name) {
-				accepted = append(accepted, name)
-			}
-		}
-		namesPerKind[names.Kind] = accepted
-	}
-	return namesPerKind
```
basti1302, line 208: Since the redaction no longer only applies to specific types, this whole part of the drift test is now meaningless.

images/agent0-connector/src/kubectl/redaction.go
```diff
@@ -112,131 +112,18 @@ var wellKnownNonSecretValues = map[string]struct{}{
 	"application/x-protobuf": {},
 }
 
-// dash0ResourceTypesWithSecrets lists the resource type names of the Dash0 custom resources whose content can contain
-// secrets: the Dash0 auth token and the header values of non-Dash0 exports (Dash0OperatorConfiguration,
-// Dash0Monitoring), the credentials of the third-party integrations of a notification channel
-// (Dash0NotificationChannel), and the credentials a synthetic check sends with its request (Dash0SyntheticCheck).
-// Singular and plural form are listed; none of these custom resources have short names, and kubectl also accepts the
-// kind (e.g. "Dash0Monitoring"), which normalizes to the singular form.
-var dash0ResourceTypesWithSecrets = map[string]struct{}{
```
basti1302, line 121: Not restricting redaction to specific resource types made all these lists obsolete.

basti1302 (commented):
.

p-d0 (approved):

PR #1406, opened 2026-09-18
feat(helm): add support for configuring http and grpc exporters

thieltges:
## Authorship

- [ ] fully understand the changes and can explain the reasoning behind the design decisions,
- [ ] fully understand how to verify and test the changes,
- [ ] did confirm the changes work as expected by testing them in a real cluster, and
- [ ] personally wrote the PR description.


helm-chart/dash0-operator/docs/installation.md
```diff
@@ -86,6 +86,38 @@ For **production use**, we strongly recommend tweaking a couple of settings:
...
+  --namespace dash0-system \
+  --create-namespace \
+  --values other-backend.yaml \
+  dash0-operator \
+  dash0-operator/dash0-operator
+```
+
+See [Exporting Data to Other Observability Backends](advanced-configuration.md#exporting-data-to-other-observability-backends)
+for the full list of settings of an `http` or `grpc` export.
+
```
basti1302: I think "Installation with a Non-Dash0 Backend" is more of an edge case and does not need to be explained in `installation.md`. The section in advanced-configuration.md should be sufficient. That is, this section can probably be removed.

helm-chart/dash0-operator/templates/operator/deployment-and-webhooks.yaml
```diff
@@ -150,13 +153,15 @@ spec:
...
 {{- end }}{{/*closes "if .Values.operator.dash0Export.keepalive.permit_without_stream" */}}
+{{- end }}{{/*closes the second "if .Values.operator.dash0Export.enabled" */}}
         - --operator-configuration-auto-monitor-namespaces-enabled={{ .Values.operator.autoMonitorNamespaces.enabled }}
         - --operator-configuration-auto-monitor-namespaces-label-selector={{ .Values.operator.autoMonitorNamespaces.labelSelector }}
 {{- if not (or (eq .Values.operator.instrumentation.delivery "auto") (eq .Values.operator.instrumentation.delivery "image-volume") (eq .Values.operator.instrumentation.delivery "init-container")) }}
 {{- fail (printf "Error: operator.instrumentation.delivery must be one of \"auto\", \"image-volume\" or \"init-container\", but was %q." .Values.operator.instrumentation.delivery) -}}
 {{- end }}
         - --operator-configuration-instrumentation-delivery={{ .Values.operator.instrumentation.delivery }}
-{{- end }}{{/*closes "if .Values.operator.dash0Export.enabled" */}}
+        - --operator-configuration-managed-via-helm=true
```
basti1302: _nit:_ let's move this to the top of its if-end block, e.g. to line 116, I think it would read a bit more naturally. (E.g. if createOperatorConfigurationResource and --operator-configuration-managed-via-helm=true directly next to each other).

helm-chart/dash0-operator/docs/advanced-configuration.md
```diff
@@ -585,6 +585,40 @@ spec:
...
+        headers:
+          - name: X-My-Header
+            value: my-value
+```
+
+Setting `operator.exports` to a non-empty list makes the operator create the operator configuration resource at startup,
+just like `operator.dash0Export.enabled=true` does; see
+[Notes on Creating the Operator Configuration Resource Via Helm](configuration.md#notes-on-creating-the-operator-configuration-resource-via-helm).
+Both can be combined: the Dash0 export configured via `operator.dash0Export.*` comes first, followed by the exports
+listed in `operator.exports`. Configuring a `dash0` export in `operator.exports` while `operator.dash0Export.enabled` is
```
basti1302: _suggestion:_ It is fine to allow combining `operator.dash0Export.enabled` with additional non-dash0 exports via `operator.exports`. But we should probably recommend to only use `operator.exports` when multiple exports are configured, and not spreading the exports over two different Helm values. 

That is, our recommendation should be that for any multi-export setup, all exports are configured via operator.exports, including the dash0 export(s) and any additional grpc or http exports.

I think it's cleaner for the user to have all exports in one place, so we should guide our users in that direction.

helm-chart/dash0-operator/values.yaml
```diff
@@ -87,6 +87,39 @@ operator:
...
       permit_without_stream:
 
+  # Additional exports for the automatically created Dash0OperatorConfiguration resource. Use this to send telemetry to
+  # an arbitrary OTLP-compatible backend via gRPC or HTTP, either instead of or in addition to Dash0. The value is a
+  # list with exactly the structure of the spec.exports list of the Dash0OperatorConfiguration resource, see
+  # https://github.com/dash0hq/dash0-operator/blob/main/helm-chart/dash0-operator/docs/advanced-configuration.md#exporting-data-to-other-observability-backends
+  #
+  # Setting this to a non-empty list makes the operator manager create the Dash0OperatorConfiguration resource at
+  # startup, exactly like setting operator.dash0Export.enabled to true does. Both can be combined; the Dash0 export
+  # configured via operator.dash0Export.* always comes first, followed by the entries listed here. It is an error to
```
basti1302: _suggestion:_ Same suggestion as for the advanced-configuration: We should recommend to only use operator.exports for all configured exports for setups that use multiple exports. (The markdown docs are often a bit redundant with the values.yaml comments, that is deliberate.)

helm-chart/dash0-operator/values.yaml
```diff
@@ -87,6 +87,39 @@ operator:
...
+  # Note that the order matters for self-monitoring: the operator's own telemetry is sent to the first export only.
+  # Self-monitoring is not supported for an HTTP export with encoding "json"; it is silently disabled in that case.
+  #
+  # Note that the values listed here are rendered verbatim into a Kubernetes ConfigMap. For header values that contain
+  # sensitive data, use valueFrom.secretKeyRef instead of a literal value; the referenced secret has to exist in the
+  # namespace the operator is installed in.
+  #
+  # Example:
+  # exports:
+  #   - grpc:
```
basti1302, line 119: suggestion: add a dash0 export here as the first export in the example.

internal/webhooks/operator_configuration_validation_webhook.go
```diff
@@ -298,9 +300,16 @@ func validateTelemetryCollectionDisabledConsistency(
...
+//
+// The Dash0SignalControl custom resource definition is only installed when the Helm value operator.signalControl.enabled
+// is true. Listing the resources then fails with a no-match error, which means no Signal Control resource can exist,
+// not that the check failed.
 func (h *OperatorConfigurationValidationWebhookHandler) hasEnabledSignalControl(ctx context.Context) (bool, string, error) {
 	allSignalControlResources := &dash0v1alpha1.Dash0SignalControlList{}
 	if err := h.Client.List(ctx, allSignalControlResources); err != nil {
+		if meta.IsNoMatchError(err) || apierrors.IsNotFound(err) {
+			return false, "", nil
+		}
```
basti1302, line 310: _praise_: Nice, that looks like it fixes a pre-existing bug! If I read this right, if a user would have manually configured a operator configuration resource without an export and without signal control enabled, this would have failed since 6bf7ebb3.

test-resources/bin/lib/util
```diff
@@ -757,7 +757,13 @@ run_helm() {
 
   # Deploy an operator configuration right away.
   if [[ "${DEPLOY_OPERATOR_CONFIGURATION_VIA_HELM:-}" != "false" ]]; then
-    if [[ "${USE_OTLP_SINK:-}" = "true" ]]; then
+    if [[ "${OPERATOR_CONFIGURATION_VIA_HELM_GRPC_EXPORT:-}" = "true" ]]; then
```
basti1302: _chore_: OPERATOR_CONFIGURATION_VIA_HELM_GRPC_EXPORT needs to be documented in CONTRIBUTING.md.

basti1302: _idea_: Instead of having OPERATOR_CONFIGURATION_VIA_HELM_GRPC_EXPORT as a boolean flag, maybe it would be nicer to have it take the value of an actual endpoint. Maybe combine that with a list of header key-value pairs in a separate env var. That way, the script could be used to actually configure an arbitrary grpc export via this new feature.

test/e2e/e2e_test.go
```diff
@@ -2871,6 +2871,60 @@ spec:
...
+					"",
+					&images,
+					// no operator.dash0Export.* values, the export is provided via operator.exports only
+					nil,
+					map[string]string{
+						"operator.exports[0].grpc.endpoint": defaultEndpoint,
+						"operator.exports[0].grpc.insecure": "true",
+						"operator.clusterName":              e2eKubernetesContext,
+						// Self-monitoring metrics are namespace-scoped, but this test verifies that only
+						// non-namespace-scoped metrics arrive, so self-monitoring is disabled deliberately.
```
basti1302: _issue_: I think self-monitoring is not supported anyway if a non-dash0 export is used, so this comment reads a little odd - it gives the wrong reason for setting selfMonitoringEnabled: false.

test/e2e/e2e_test.go
```diff
@@ -2871,6 +2871,60 @@ spec:
...
+					g.Expect(*exports[0].Grpc.Insecure).To(BeTrue())
+				}, 30*time.Second, pollingInterval).Should(Succeed())
+
+				By("waiting for metrics")
+				Eventually(func(g Gomega) {
+					verifyNonNamespaceScopedKubeletStatsMetricsOnly(g, timestampLowerBound)
+				}, 50*time.Second, time.Second).Should(Succeed())
+			})
+		})
+
```
basti1302, line 2928: _praise:_ Good call to include a new test-resources/bin script and an e2e test. 👍 👍

internal/webhooks/operator_configuration_validation_webhook.go
```diff
@@ -298,9 +300,16 @@ func validateTelemetryCollectionDisabledConsistency(
 // hasEnabledSignalControl reports whether an enabled Dash0SignalControl resource exists in the cluster. It returns the
 // name of the first such resource for logging. Signal Control requires a Dash0 export, so the operator configuration
 // must not drop its Dash0 export while Signal Control is enabled.
+//
+// The Dash0SignalControl custom resource definition is only installed when the Helm value operator.signalControl.enabled
+// is true. Listing the resources then fails with a no-match error, which means no Signal Control resource can exist,
+// not that the check failed.
```
basti1302: _nit_: I think this is not relevant for the godoc of the hasEnabledSignalControl, it is the typical AI-generated slop of documenting the motivation of a particular code change in the comments.

Suggestion: Move it down to line 311, as an implementation comment explaining why NoMatch and NotFound are treated separately. There it would make a lot of sense as an explanation.

basti1302 (commented):
Good work! 🎉  Very complete. I left a couple of comments inline - nothing major though.

test/e2e/e2e_test.go
thieltges: i think self-monitoring does work with non-dash0 export. the operator manager sends it to grpc/http export and the collectors self-monitoring pipeline uses the same export conversion. but http with encoding: json won't work. so with self-monitoring on, the operator's namespaced metrics would land in the sink and break the "non-namespaced metrics only" assertion. i just reworded the comment to make that clearer. or am i missing something?

basti1302: You are right, I misremembered that. Nevermind 👍

basti1302 (approved):
LGTM 👍 🎉

